PR #39, opened 2023-06-28
Uasc 20/member bookings

mwon369:
Completed:

- Created a booking form that currently contains an MUI date-picker component which has some logic for checking valid user input.

Next Steps:

- Connect the Booking Form to the Firebase backend so that bookings alongside the user's information can be added to Firestore when the form is submitted.

Notes:

- Didn't add input fields on the booking form for the user to enter in their personal details, since I figured we could just grab those details from the user object we get from Fireauth since the user should be logged in to fill out the form in the first place - could be wrong though.


BW82262 (approved):
LGTM

mwon369:
Wait hol up don't merge yet

PR #40, opened 2023-07-03
Uasc 35/auto hide navbar on home page

mwon369:
Completed:

- Navbar is now sticky so it remains in view as the user continues to scroll down a page.

- Navbar is hidden by default when visiting the home page and is only displayed when the user has scrolled more than a certain proportion (currently set to 30% but can be easily changed) of the view height. The navbar also hides again if the user scrolls back up to the top of the home page.

Notes:

- This behaviour is only present on the home page - on every other page the navbar is shown by default and doesn't show/hide on scroll.


BW82262 (approved):
LGTM, might need to revisit this later when the homepage is done.

Issue #24, opened 2023-05-01
Setup Firebase Emulator


OverHash:
### User story:

As a **developer**, I want to **have a local emulator of Firebase**, so that **I don't use the production database when developing**.

### Description:
We have a few incentives to setup the local emulator:
- We don't want to read/write data to our production database in case something goes wrong
- We may want to have more control over what the state of the Firebase app looks like during development
- Reduce the usage of the real stores to avoid billing costs
- Allow potential for unit testing of our app

### Acceptance Criteria:
- [ ] Investigated Firebase Emulator as a viable tool to add to the development pipeline
- [ ] Setup Firebase Emulator on the Firebase end
- [ ] Connected Firebase Emulator to React app

### Definition of Done:
- [ ] AC met
- [ ] Production app will still read production data
- [ ] Local app will use Firebase emulator
- [ ] Applicable documentation is written in GH wiki
- [ ] PR reviewed
- [ ] Merged into main/feature branch

### Notes / Links
Read more about the Firebase emulator suite [here](https://firebase.google.com/docs/emulator-suite)

PR #41, opened 2023-07-04
Fix production build of emulator suite

OverHash:
We hook into the native feature of create-react-app which sets the `NODE_ENV` environment variable to be `development` when running `npm start`, and `production` when running `npm run build` (as we do when deploying).

This allows us to toggle between using the emulator suite, or the production store.


OverHash:
Closes #24

BW82262 (approved):
LGTM

PR #47, opened 2023-07-09
Uasc 38/profile data

tyin363:
Completed:
- Get current user data from database (one user only)

Next Steps:
- Fetch correct user data when user authentication system is finished


BW82262 (approved):
LGTM

PR #49, opened 2023-07-10
UASC-28/Member-Login

atharvaarankalle:
Completed:

- Form made using Material UI. Two fields: Email and Password
- On submit, a success/error message is displayed for two seconds depending on if the login was successful or not. If the login is successful, user is redirected to the profile page
- Tested using Firebase emulator

Next Steps:

- Possibly pass the information of the logged in user to the profiles page for further use
- Styling of the login form and login page


OverHash:
Should we handle the case of `auth/invalid-email` ([error codes](https://firebase.google.com/docs/reference/node/firebase.auth.Auth#error-codes_11))

frontend/src/components/LoginForm.js
```diff
@@ -0,0 +1,65 @@
...
+            .catch((error) => {
+                switch (error.code) {
+                    case "auth/wrong-password":
+                        setLoginErrorMessage("Incorrect password");
+                        setTimeout(() => {
+                            setLoginErrorMessage("");
+                        }, 2000);
+                        break;
+                    case "auth/user-not-found":
+                        setLoginErrorMessage("User not found");
```
OverHash, line 35: "User" is quite a technical term. We should probably word this something like "Email does not have an account associated with it".

Sidenote: I don't like how Firebase exposes the email like this -- it allows an arbitrary user to see if someone has signed up with a certain email account.

BW82262 (approved):
lgtm

PR #57, opened 2023-07-17
Uasc 33/navbar display necessary tabs

selinakkaya:
Code for navbar and code for displaying the necessary tabs in the nav bar.

### DIFF
--- a/frontend/src/components/Navbar.js
+++ b/frontend/src/components/Navbar.js
@@ -25,21 +25,21 @@ const Navbar = () => {
               <a href="/home">Home</a>
             </li>
             <li>
-              <a href="#about">About</a>
+              <a href="/about">About</a>
             </li>
             <li>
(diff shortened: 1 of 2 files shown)


frontend/src/components/Navbar.js
```diff
@@ -1,25 +1,62 @@
...
+        <div id="one">
+          <img src="https://uasc.co.nz/wp-content/uploads/2021/05/UASC-LOGO-White.png" alt="logo"/>
+        </div>
+        <div id="two">
+          <ul>
+            <li>
+              <a href="/home">Home</a>
+            </li>
+            <li>
+              <a href="#about">About</a>
```
BW82262, line 28:
```suggestion
              <a href="/about">About</a>
```
The href for the tag should contain a '/' rather than a '#'

BW82262 (approved):

PR #68, opened 2023-08-02
Uasc 43/showing available bookings to members

mwon369:
Completed:

Created a basic booking form that uses MUI's date picker component. Dates in the date picker are automatically disabled if there are no available spots left. Next to the date picker there is also a list which displays the users currently selected booked dates (from start date all the way to but not including the end date) alongside how many spots there are left for each date.

Note:

- There is currently placeholder code on lines 144 to 153,
- I set the maxSpotsAvailablePerDay = 5 for testing purposes (can change it to 50 in a later PR)
- I didn't manually test the functionality for displaying the number of booked spots too extensively since I couldn't get the Firestore pre-production to work this morning - here's the error: 
"Uncaught (in promise) FirebaseError: 
false for 'list' @ L6"

Next Steps:

- Rewrite any placeholder/test code to be the actual implementation so that the UI dynamically updates
- Change the submit function to post the new booking to Firestore


BW82262 (approved):
Seems to be working now. LGTM

OverHash:
Looks good for now. In the future we will want to update the fetch part of this to send a [query](https://firebase.google.com/docs/firestore/query-data/queries) between the date range specified to only fetch data we need.

PR #70, opened 2023-08-02
uasc-44/admin-dashboard-layout added RequestBlock component

svimi:
ive created a request block component to render any data we want to fetch from firebase. at the moment it only works with dummy data manually put in. potential changes in the future can include scroll bar stylization for the request list, among other things.


frontend/src/components/AdminRequests.js
```diff
@@ -1,220 +1,90 @@
...
-                      DATE CHANGE
-                      <span style={textBreakStyle}></span>
-                      5/7 - 7/7 -&gt; 7/7 - 9/7
-                    </Typography>
-                  </Box>
-                </Paper>
-              </Grid>
+              {requests.map((request, index) => (
+                <RequestBlock
+                  key={request.user_id}
```
OverHash, line 83: A user ID isn't a unique key (a user can have multiple requests). We should use the request document ID returned by Firebase, something like
```js
key={request._id}
```

frontend/src/components/RequestDetails.js
BW82262, line 1: I'm assuming this is used for your next user story?

BW82262 (approved):
lgtm

frontend/src/components/AdminRequests.js
```diff
@@ -22,8 +22,8 @@ const AdminDashboardRequest = () => {
         {
           user_id: "whatsgood",
           booking_id: "8mYj7rWOMH6hGy4FzMed",
-          query: "I'd like to cancel my booking.",
-          query_type: "cancellation",
+          query: "I'd like to change my booking.",
+          query_type: "date change",
```
OverHash, line 26:
```suggestion
          query_type: "dateChange",
```

OverHash (changes requested):

frontend/src/components/AdminRequests.js
svimi: changed

frontend/src/components/RequestDetails.js
svimi: yea

frontend/src/components/AdminRequests.js
svimi: is it okay to just use the booking id? thats what ive set it as right now

OverHash (changes requested):
Please delete `node_modules` folder in root and the root `package.json`.

PR #73, opened 2023-08-09
Add persistent user data hook

OverHash:
Adds a new `useAuthenticatedUser` hook which returns both the Firebase user object and the associated user collection metadata document for the currently authenticated user.


BW82262 (approved):
lgtm

PR #74, opened 2023-08-10
Uasc 51/bookings calendar overview

edl21:


BW82262 (approved):
Functionality looks good. Think the next steps would be to move the bookings code with the calendar overview to its own separate component.

PR #75, opened 2023-08-15
Uasc 48/style profile page

BW82262:
Completed:
The user profile page is styled up according to the new design.

Note:
The data is currently static and will need to be dynamically updated later on.


mwon369 (approved):
LGTM!

PR #95, opened 2023-09-06
Extend Firestore rules all-read access

OverHash:
Extends read access to the Firestore up until October 22th. This should give us enough time until we deploy proper read rules.

### DIFF
--- a/frontend/firestore.rules
+++ b/frontend/firestore.rules
@@ -3,7 +3,7 @@ service cloud.firestore {
   match /databases/{database}/documents {
     match /{document=**} {
       allow read, write: if
-          request.time < timestamp.date(2023, 8, 31);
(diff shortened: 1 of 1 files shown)


frontend/firestore.rules
```diff
@@ -3,7 +3,7 @@ service cloud.firestore {
   match /databases/{database}/documents {
     match /{document=**} {
       allow read, write: if
-          request.time < timestamp.date(2023, 8, 31);
+          request.time < timestamp.date(2023, 09, 22);
```
OverHash, line 6: It should be noted that we don't suffer from [zero prefixed octal numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Numbers_and_dates) in JavaScript here as `9` is not an octal digit.

BW82262 (approved):

PR #99, opened 2023-09-16
Uasc 71/booking request UI functionality

BW82262:
Added Functionality that allows users to make a request for their bookings.
Currently only permits for dateChange request for when the total days remain the same.
Will create separate story for cancellation functionality as it requires a UI.


mwon369 (approved):
LGTM

Issue #82, opened 2023-08-16
User Profile Page -Display Profile Data


BW82262:
### User story:

As a user,
I want to see my account details,
So that I can see if they're correct.

### Acceptance Criteria:
- [ ] User Data is obtained from DB
- [ ] Data is displayed on profile

### Definition of Done:
- [ ] AC met
- [ ] Applicable tests written
- [ ] Applicable documentation is written in GH wiki
- [ ] PR reviewed
- [ ] Merged into main/feature branch

PR #100, opened 2023-09-16
82/display user profile data

BW82262:
Issue #82 
Displays user details on profile page. 
Currently shows the essential fields. 
Emergency contacts and such will need to be added later as separate field. Not part of the sign up details.


mwon369 (approved):
LGTM

Issue #102, opened 2023-09-17
Fix create bookings functionality


OverHash:
The current [create bookings](https://github.com/UoaWDCC/uasc-web/blob/f03b080cb13b6d082192f355394927fc16dc3f35/frontend/src/components/BookingForm.js) functionality suffers from three issues:
1. The fields are not added in accordance with [the ARCHITECTURE.md](https://github.com/UoaWDCC/uasc-web/blob/f03b080cb13b6d082192f355394927fc16dc3f35/frontend/ARCHITECTURE.md#bookings-collection) definition of how the bookings should be inserted (namely `checkIn` is inserted instead of `check_in`, and a similar casing issue exists with `checkOut`)
2. The document inserts the booking under a `uid` field (should be `user_id`), but does not insert it as a reference to the actual authenticated user (using the [useAuthenticatedUser](https://github.com/UoaWDCC/uasc-web/blob/f03b080cb13b6d082192f355394927fc16dc3f35/frontend/src/hooks/useAuthenticatedUser.js) hook)
3. The inserted document dates are not normalized to be at midnight. We should make this an informal convention (or update the `ARCHITECTURE.md`) to specify this to avoid having weird bugs in the future.

### Acceptance Criteria:
- [ ] Fields are specified in accordance with the [the ARCHITECTURE.md](https://github.com/UoaWDCC/uasc-web/blob/f03b080cb13b6d082192f355394927fc16dc3f35/frontend/ARCHITECTURE.md#bookings-collection) definition
- [ ] Created booking correctly references the authenticated user
- [ ] Dates are normalized to use midnight as the time

### Definition of Done:
- [ ] AC met
- [ ] PR reviewed
- [ ] Merged into main/feature branch

Issue #101, opened 2023-09-17
Fix navbar login/logout functionality


OverHash:
Currently the [navbar](https://github.com/UoaWDCC/uasc-web/blob/f03b080cb13b6d082192f355394927fc16dc3f35/frontend/src/components/Navbar.js) defaults to a logged in state (displaying "logout"). We should fix this functionality to properly use the [useAuthenticatedUser](https://github.com/UoaWDCC/uasc-web/blob/main/frontend/src/hooks/useAuthenticatedUser.js) hook to display the information correctly.

### Acceptance Criteria:
- [ ] navbar shows "login" when user is actually logged out
- [ ] navbar shows "logout" when user is logged out
- [ ] navbar state saves if user closes tab and comes back
- [ ] navbar actually calls firebase auth logout when logging out

### Definition of Done:
- [ ] AC met
- [ ] PR reviewed
- [ ] Merged into main/feature branch

PR #103, opened 2023-09-17
Uasc 90/authenticated user details

### LINKED ISSUE #90: User Booking - Authenticated User Details
### User story:

As a member,
I want to make a lodge booking,
So that I have a spot at the lodge for the time I booked.

### Description:
Uses the useAuthenciatedUser hook from the home page to get user details
This assumes the member's membership is VALID.

### Acceptance Criteria:
- [ ] Users who are not authenticated can't make bookings, but can view
- [ ] Uses the user detail from hook to making booking

### Definition of Done:
- [ ] AC met
- [ ] Applicable tests written
- [ ] Applicable documentation is written in GH wiki
- [ ] PR reviewed
- [ ] Merged into main/feature branch

mwon369:
closes #90

### DIFF
--- a/frontend/src/components/BookingForm.js
+++ b/frontend/src/components/BookingForm.js
@@ -23,12 +23,10 @@ const BookingForm = () => {
   const [user, userMetadata] = useAuthenticatedUser()
   const [userData, setUserData] = useState(undefined)
 
-  // fetch bookings on component mount
   useEffect(() => {
     retrieveExistingBookings()
   }, [])
 
-  // build hashmap of booked spot availability once existingBookings state has been set
   useEffect(() => {
     setSpotAvailabilityByDate()
   }, [existingBookings])
@@ -43,7 +41,7 @@ const BookingForm = () => {
 
   useEffect(() => {
     setUserData(userMetadata)
-  }, [user, userMetadata])
+  }, [user])
 
   const retrieveExistingBookings = async () => {
     const querySnapshot = await getDocs(bookingCollectionRef)
--- a/frontend/src/components/BookingForm.js
+++ b/frontend/src/components/BookingForm.js
@@ -41,7 +41,7 @@ const BookingForm = () => {
 
   useEffect(() => {
     setUserData(userMetadata)
-  }, [user])
+  }, [user, userMetadata])
 
   const retrieveExistingBookings = async () => {
     const querySnapshot = await getDocs(bookingCollectionRef)
--- a/frontend/src/components/BookingForm.js
+++ b/frontend/src/components/BookingForm.js
@@ -4,17 +4,24 @@ import { DatePicker } from "@mui/x-date-pickers"
 import dayjs from "dayjs"
 import { db } from "../firebase"
 import { getDocs, collection, addDoc } from "firebase/firestore"
+import { useAuthenticatedUser } from "../hooks/useAuthenticatedUser"
 
 const BookingForm = () => {
   const [selectedCheckInDate, setSelectedCheckInDate] = useState(null)
   const [selectedCheckOutDate, setSelectedCheckOutDate] = useState(null)
+
   const [existingBookings, setExistingBookings] = useState([])
   const [dateAvailabilities, setDateAvailabilities] = useState(new Map())
   const [dateRange, setDateRange] = useState([])
+
   const bookingCollectionRef = collection(db, "bookings")
   const maxSpotsAvailablePerDay = 50
+
   const [bookingSuccessful, setBookingSuccessful] = useState(false)
   const [bookingErrorMessage, setBookingErrorMessage] = useState("")
+
+  const [user, userMetadata] = useAuthenticatedUser()
+  const [userData, setUserData] = useState(undefined)
 
   // fetch bookings on component mount
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/BookingForm.js
```diff
@@ -34,6 +39,10 @@ const BookingForm = () => {
     setDateRange(getDateRangeArray())
   }, [selectedCheckInDate, selectedCheckOutDate])
 
+  useEffect(() => {
+    setUserData(userMetadata)
+  }, [user, userMetadata])
```
OverHash, line 44:
```suggestion
```
We should just rename `userMetadata` to be `userData` here.

frontend/src/components/BookingForm.js
```diff
@@ -4,24 +4,29 @@ import { DatePicker } from "@mui/x-date-pickers"
...
+
   const bookingCollectionRef = collection(db, "bookings")
   const maxSpotsAvailablePerDay = 50
+
   const [bookingSuccessful, setBookingSuccessful] = useState(false)
   const [bookingErrorMessage, setBookingErrorMessage] = useState("")
 
-  // fetch bookings on component mount
+  const [user, userMetadata] = useAuthenticatedUser()
+  const [userData, setUserData] = useState(undefined)
```
OverHash, line 24:
```suggestion
```

frontend/src/components/BookingForm.js
```diff
@@ -4,24 +4,29 @@ import { DatePicker } from "@mui/x-date-pickers"
...
   const [dateRange, setDateRange] = useState([])
+
   const bookingCollectionRef = collection(db, "bookings")
   const maxSpotsAvailablePerDay = 50
+
   const [bookingSuccessful, setBookingSuccessful] = useState(false)
   const [bookingErrorMessage, setBookingErrorMessage] = useState("")
 
-  // fetch bookings on component mount
+  const [user, userMetadata] = useAuthenticatedUser()
```
OverHash, line 23:
```suggestion
  const [user, userData] = useAuthenticatedUser()
```

frontend/src/components/BookingForm.js
```diff
@@ -50,8 +59,8 @@ const BookingForm = () => {
     let availabilities = new Map()
 
     existingBookings.forEach((booking) => {
-      let currDate = new Date(getFormattedDateString(booking.checkIn))
-      const endDate = new Date(getFormattedDateString(booking.checkOut))
+      let currDate = new Date(getFormattedDateString(booking.check_in))
```
OverHash, line 62: Unrelated to this PR but do we need dayjs? The `booking.check_in` is a Firestore [timestamp](https://firebase.google.com/docs/reference/js/v8/firebase.firestore.Timestamp) which has a `toDate` method itself. Just a comment for other maintainers.

frontend/src/components/BookingForm.js
BW82262: Yea, probably. I initially thought that user was the actual user data itself.

frontend/src/hooks/useAuthenticatedUser.js
```diff
@@ -19,9 +19,7 @@ export function useAuthenticatedUser() {
       }
     }
 
-    if (user) {
-      getUserData()
-    }
+    user ? getUserData() : setUserData(null)
```
OverHash, line 22: Any reason we want to use `null` instead of `undefined` here? What was wrong with the old behavior of leaving it as `undefined`?

OverHash (changes requested):
Overall looks pretty good. Just a few small tidbits

frontend/src/components/BookingForm.js
OverHash: Yeah this is an unfortunate side-effect of the upstream hook we're using. I considered merging the two objects to be one, but settled for having it as a "metadata" part. Perhaps we can re-look that if it's still confusing.

OverHash:
Looks like this fulfills the last item of #101.

It will also fix the first two parts of #102.

frontend/src/components/BookingForm.js
mwon369: Yeah I was thinking this too. I only used that useEffect so that there was a reason to use the "user" variable since ESLint would get mad about unused variables. My smooth brain didn't know I could've just done `[, userData] = useAuthenticatedUser()` to destructure without having a user variable

frontend/src/hooks/useAuthenticatedUser.js
mwon369: Oh yeah I added this just to explicitly set the userMetaData to something falsey whenever the auth state (`[user]`) was changed. This was done so that whenever the user clicked logout, the userMetaData would update to reflect the logout change in `[user]`, since the meta data is being used for posting the booking to Firestore.

frontend/src/components/BookingForm.js
mwon369: Iirc, I think I added it in because I was getting some runtime errors due to incompatible types between the booking form state and the selectable dates of the MUI date picker.

But if there is a better way to do it, I think it would be a good idea to look into refactoring a little bit to get rid of dayjs if it's not absolutely needed.

BW82262 (approved):

frontend/src/components/BookingForm.js
OverHash: If mui depends on it, then we can just leave it as is.

frontend/src/hooks/useAuthenticatedUser.js
OverHash: Gotcha (good catch!). We should probably stick to `userData` being either exactly `undefined` or the user, or exactly `null` or the user (i.e., not default to `undefined` and then later have the potential to be `null`).

I think either is fine, I'd probably lean to `undefined` though.

mwon369: Sounds good to me, I've made another commit to set the userMetaData to be undefined instead of null on sign out 👍

OverHash (approved):
LGTM :rocket:!

PR #105, opened 2023-09-17
Display user bookings in profile page

OverHash:
We now display the user bookings on the `/profile` page, which shows:
1. a calendar overview with highlighted dates they have booked
2. a list of all their current bookings (alongside the request changes button)

image

### DIFF
--- a/frontend/src/pages/Profile.js
+++ b/frontend/src/pages/Profile.js
@@ -1,6 +1,6 @@
 import { db } from "../firebase"
 import React, { useEffect, useState } from "react"
-import { getDocs, where, query, collection } from "firebase/firestore"
+import { getDocs, where, query, collection, doc } from "firebase/firestore"
 import { Stack, Typography } from "@mui/material"
 import "../styles/Profile.css"
 import ProfileCard from "../components/ProfileCard"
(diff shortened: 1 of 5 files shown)


frontend/src/pages/Profile.js
```diff
@@ -9,36 +9,45 @@ import ProfileCurrentBookings from "../components/ProfileCurrentBookings"
...
+
+  /**
+   * Retrieves the current user bookings from firestore.
+   * @param {string} userId
+   */
+  const getBookings = (userId) => {
+    getDocs(
+      query(
+        collection(db, "bookings"),
+        where("user_id", "==", doc(db, "users", userId))
```
BW82262, line 32: What was the reason that `doc(db, "users", userId)` was used instead of just `userId`? 
It doesn't seem to be able to get any of the collection data with this change for me.

OverHash: Hmm -- that's interesting. Testing locally with the emulator doesn't work unless I do it in this way.

My understanding is that using `doc(db, "users", userId)` here will make the query against a [DocumentReference[(https://firebase.google.com/docs/reference/js/v8/firebase.firestore.DocumentReference) (i.e., the `/users/userId` reference) rather than a string literal of the user id.

BW82262 (approved):
LGTM!

PR #112, opened 2023-09-24
Stripe webhook setup + Thanks page

TFPlusPlus:
The Stripe webhook is set up and can be tested using express (the live online endpoint itself will have to be set up on Firebase) A Thanks page is ready for redirecting users from the Stripe payment page.


TFPlusPlus:
Also need to take care of sending the event to the database to record the payment.

PR #113, opened 2023-10-02
uasc-92/UserBooking-BookingHistory

edl21:
There's an issue with the firestore.rules at the moment with the date being 9/30 in this version of the main. Apart from that, functionality should be working properly.

### DIFF
--- a/frontend/src/components/ProfileBookingHistory.js
+++ b/frontend/src/components/ProfileBookingHistory.js
@@ -1,64 +1,120 @@
-import React from "react"
-import { CardContent, Stack, Card, Typography } from "@mui/material"
+import React, { useState, useEffect } from "react"
+import { db } from "../firebase"
+import {
+  Table,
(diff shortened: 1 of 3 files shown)


frontend/src/components/ProfileBookingHistory.js
```diff
@@ -1,64 +1,120 @@
...
+          id: doc.id,
+          ...doc.data(),
+        }))
+        setBookings(fetchedBookings)
+      } catch (error) {
+        console.error("Error fetching user bookings: ", error)
+      } finally {
+        setLoading(false)
+      }
+    }
```
BW82262, line 54: It seems like the bookings data is being fetch here within the component. The booking data is already fetched in profile and could be used instead.

BW82262 (commented):
The current method that the booking history is shown seems to display all the bookings, could this be changed to show the ones past the current date?

PR #191, opened 2024-03-28
Allows stripe webhook receiving. 

jeffplays2005:
Allows stripe webhook receival. Open to new services being added.

Updated README.md to describe how local testing can be performed on the 
Stripe webhook. 

Updated index.ts to allow for the keeping of the raw body during 
requests. Plans to change this to only enable for the /webhook endpoint.


server/src/service-layer/controllers/README.md
```diff
@@ -3,3 +3,9 @@
 Controller methods convert requests coming from the routes and convert to HTTP responses
 
 See the [tsoa docs](https://tsoa-community.github.io/docs/examples.html) for how to write controllers (These will automatically be converted to routes)
+
+## Stripe Webhook
```
bcho892, line 7: Great job! nice to see documentation

nit: This would be better in the wiki i think

server/src/business-layer/services/StripeWebhook.test.ts
bcho892, line 1: these tests should be reserved for services

server/src/index.ts
```diff
@@ -15,8 +15,19 @@ const importSwaggerJson = async () => {
 
 const app: Express = express()
 
+function keepRawBody(
```
bcho892, line 18: 👍 

nit: can we extract this out of `index.ts`?

client/package.json
```diff
@@ -33,7 +33,7 @@
     "react-router-dom": "^6.10.0",
     "react-sweet-state": "^2.7.1",
     "scheduler": "^0.23.0",
-    "stripe": "^13.7.0",
+    "stripe": "^14.22.0",
```
bcho892, line 36: this should be installed in the server workspace instead?

bcho892 (changes requested):

bcho892:
Well done on dealing with all the errors haha, most of this looks good to me, just some changes needed so we aren't making massive changes to the repo

server/src/business-layer/services/StripeWebhook.test.ts
jeffplays2005: Removed the test!

server/src/service-layer/controllers/README.md
jeffplays2005: Moved to wiki.

client/package.json
jeffplays2005: Reverted back to master. Updated server workspace.

PR #198, opened 2024-03-29
172-Annotate-UserAdditionalInfo

Kartik-M24:
The model has been annotated, just wanted to double check it is done properly.


bcho892:
this one is cooked

PR #199, opened 2024-03-29
Added annotations and missing fields to UserAdditionalInfo

Kartik-M24:
Clarification on whether task is complete


server/src/data-layer/models/firebase.ts
```diff
@@ -1,22 +1,74 @@
...
+   */
   does_racing: boolean
+  /**
+   * @isbool Please enter true or false
+   */
   does_ski: boolean
+  /**
+   * @isString Please enter your pronouns
+   */
+  gender: string
```
bcho892, line 23: since you added a new field here, you will need to update the user mock (`User.mock.ts`) so the tests will pass

bcho892 (changes requested):
lgtm, just gotta fix the tests and run both the code generation commands (`yarn workspace client generate-types` and `yarn workspace server tsoa spec-and-routes`)

bcho892 (approved):
👍

PR #203, opened 2024-03-30
store current user in sweet state

### LINKED ISSUE #141: [FRONTEND] Add logic to store current user data in sweet state
**Is your feature request related to a problem? Please describe.**
We want information about the current user to be persisted across the application so we can display the correct information

**Describe the solution you'd like**
In our firebase config we should save user data in sweet state.

For this, we will also have to create an [action](https://atlassian.github.io/react-sweet-state/#/basics/actions) for both getting the user data and setting it. call the setter inside a [firebase auth callback](https://stackoverflow.com/questions/67785025/dispatch-action-outside-component-in-react-sweet-state)

**BEFORE MERGING**
- [ ] Storybooks created where possible
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

bcho892:
closes #141 

We now have our user data stored in sweet state.

This is similar to the work @8cw did in the `useAuthenticatedUser` hook, but this one gets triggered on the id token changing for the user in auth

There are two objects stored: `currentUser`, which is a reference to the user object from the firebase auth and `currentUserData`, which is of type `UserAdditionalInfo` (which is similar to the `userMetaData` in the old hook)

to use this user in components, use `const [{currentUser, currentUserData, currentUserClaims}] = useAppData()` (you can exclude properties you aren't using)

from there you can call functions like: `await currentUser.getIdToken(true)` to refresh the user. Or access the information/claims by using `currentUserData`/`currentUserClaims`

### DIFF
--- a/client/src/firebase.ts
+++ b/client/src/firebase.ts
@@ -2,6 +2,7 @@
 import { initializeApp, type FirebaseOptions } from "@firebase/app"
 import { getAuth, connectAuthEmulator } from "@firebase/auth"
 import { getFirestore, connectFirestoreEmulator } from "@firebase/firestore"
+import fetchClient from "services/OpenApiFetchClient"
 import { StoreInstance } from "store/store"
 
 const firebaseConfig: FirebaseOptions = {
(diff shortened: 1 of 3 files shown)


asun555 (approved):
lgtm

PR #205, opened 2024-03-31
Store jwt token in memory every time it refreshes

bcho892:
image

It is added to requests now as shown

Also added some error handling to the other network calls in the callback


asun555 (approved):
LGTM

client/src/firebase.ts
```diff
@@ -22,7 +22,6 @@ const db = getFirestore(app)
 // use emulator suite if running locally
 if (import.meta.env.VITE_NODE_ENV !== "production") {
   connectFirestoreEmulator(db, "localhost", 8080)
-  connectAuthEmulator(auth, "http://localhost:9099")
```
bcho892, line 25: we may have to readd this once we start introducing jest

PR #207, opened 2024-04-02
137 auth service functionality

ganznz:


server/src/business-layer/services/AuthService.test.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+
+describe("AuthService Unit Tests", () => {
+  it("creates a user", async () => {
+    const createdUser = await new AuthService().createUser("test@gmail.com")
+    let user
+    try {
+      user = await auth.getUserByEmail("test@gmail.com")
+    } catch {}
+    expect(createdUser).toEqual(user)
```
bcho892, line 23: maybe also add an assertion to see if the email is correct?

server/src/business-layer/services/AuthService.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    new AuthService().deleteUser("test")
+    let user
+    try {
+      user = await auth.getUser("test")
+    } catch {}
+    expect(user).toBe(undefined)
+  })
+})
+
+describe("AuthService Unit Tests", () => {
```
bcho892, line 16: nit: everything should be in the same test suite (describe block)

bcho892 (commented):
Looks good, also I remembered what the custom claims meant in the ticket, it was to add functionality for adding claims given a uid, I can implement this if you want a break, but its pretty similar to what you've done here

server/src/business-layer/services/AuthService.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  public async setCustomUserClaim(
+    uid: string,
+    role: typeof AuthServiceClaims.MEMBER | typeof AuthServiceClaims.ADMIN
+  ) {
+    let userRecord: UserRecord
+    try {
+      userRecord = await auth.getUser(uid)
+      auth.setCustomUserClaims(userRecord.uid, { [role]: true })
```
bcho892, line 44: nit: we should have the ability to remove the claim as well

bcho892 (approved):

PR #208, opened 2024-04-02
154 modularise routes

ganznz:


server/src/business-layer/services/AuthService.test.ts
bcho892, line 1: should remove this file from the pr with `git checkout master server/src/business-layer/services/AuthService.test.ts`

server/src/business-layer/services/AuthService.ts
bcho892, line 1: Should be removed `git checkout master server/src/business-layer/services/AuthService.ts`

server/src/business-layer/utils/AuthServiceClaims.ts
bcho892, line 1: should be removed `git checkout master server/src/business-layer/utils/AuthServiceClaims.ts`

bcho892 (changes requested):

bcho892 (approved):

PR #210, opened 2024-04-03
Add volta to repository

8cw:
Adds [https://volta.sh/](https://volta.sh/) configuration to package.json as an alternative to nvm.

Volta is used by other repositories such as TypeScript to manage the toolchain of package managers.

We could consider changing the onboarding flow for future contributors to simply be:
- install Volta with something akin to `curl https://get.volta.sh | bash`
- now both `node` and `yarn` are installed, and enforced to be the same version across everyone who contributes

This would remove our need to have people go through corepack to install yarn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,5 +67,9 @@
       "<rootDir>/client/jest.config.client.ts",
       "<rootDir>/server/jest.config.server.ts"
     ]
+  },
+  "volta": {
+    "node": "20.0.0",
+    "yarn": "4.1.1"
   }
 }


bcho892 (approved):
👍 would we still need to specify `packageManager`? I think this is a good change, if possible are you up to update the onboarding wiki page?

bcho892:
tested it out myself and its pretty easy to get working

8cw:
Happy to make the changes, but removing `packageManager` would force all current contributors to switch to volta.

To do so, contributors would have to `rm -rf $NVM_DIR ~/.npm ~/.bower`, then remove `nvm` from whichever PATH profile it was placed in(e.g., `~/.bash_profile`, `~/.zsh`, `~/.profile`, ...).

Then they could install volta and everything should work fine.

So I'm not sure if it's worth the complexity of forcing everyone to switch over to volta. Perhaps it's best to leave it with both methods for now.

PR #217, opened 2024-04-03
128 reusable stripe embedded payment component

zlrkw11:


client/src/services/Stripe.ts
```diff
@@ -0,0 +1,3 @@
+import { loadStripe } from "@stripe/stripe-js"
+
+export const stripePromise = loadStripe(import.meta.env.VITE_ENV_STRIPE_API_KEY)
```
bcho892, line 3: Can we rename this to `VITE_ENV_STRIPE_PUBLISHABLE_KEY` instead?

bcho892 (changes requested):

client/src/services/Stripe.ts
zlrkw11: done

bcho892 (approved):

PR #218, opened 2024-04-04
Endpoint for admins to update user membership

jeffplays2005:
Created two endpoints, "promote" and "demote", both takes a PUT request. 
Checks for forbidden requests(is admin) and conflicting requests(already  promoted/demoted).
Updates UserService and AuthService accordingly.  
AuthService updated to remove claims when "null" is passed.


server/src/service-layer/controllers/UserController.ts
```diff
@@ -75,4 +79,62 @@ export class UsersController extends Controller {
...
+    @Body() requestBody: promoteUserRequestBody
+  ): Promise<void> {
+    const userService = new UserDataService() // create a new data service
+    const authService = new AuthService()
+    // attempt to fetch user
+    const user: UserAdditionalInfo = await userService.getUserData(
+      requestBody.uid
+    )
+    if (!user) return this.setStatus(400) // bad request
+    if (user.membership === "admin") return this.setStatus(403) // admin forbidden
```
bcho892, line 98: we should not be checking this in firestore database, we need to check the `customClaims option` on the user. Should add a method in `AuthService` thats uses `auth.getUser(uid)`, the `userRecord` returned will have a `customClaims` property

bcho892 (commented):
looks good, just need to not rely on firestore for user roles

server/src/service-layer/controllers/UserController.ts
```diff
@@ -75,4 +79,62 @@ export class UsersController extends Controller {
...
+    if (user.membership === "member") return this.setStatus(409) // conflict
+    try {
+      // update user in UserService
+      await userService.editUserData(requestBody.uid, {
+        membership: "member"
```
bcho892, line 103: hm, we might have to update `UserAdditionalInfo` to allow `undefined`

server/src/business-layer/services/AuthService.test.ts
bcho892, line 1: 👍

bcho892 (commented):
LGTM, just need some tests

bcho892 (approved):
👍 i just cleaned up to avoid side effects by properly cleaning up after each test

jeffplays2005:
thank you so much for the commits benson!

PR #220, opened 2024-04-04
214 figma buttons

zlrkw11:


bcho892 (changes requested):
The disabled vairant should be handled with the `disabled:` in css, not js.

bcho892 (approved):

PR #234, opened 2024-04-06
132-stripe-webhook-logic draft pr

jeffplays2005:
Changed StripeWebhook endpoint. 
Added metadata model


server/src/business-layer/utils/StripeService.ts
bcho892, line 1: I think we should define the `CHECKOUT_TYPE_KEY` and `CHECKOUT_TYPE_VALUES`

PR #253, opened 2024-04-10
245 stepper component

zlrkw11:


client/src/components/generic/StepperComponent/StepperComponent.story.tsx
bcho892, line 1: I'd like to see an example where the text foir each step are different lengths

bcho892 (changes requested):

client/src/components/generic/StepperComponent/StepperComponentLongText.story.tsx
bcho892, line 1: This should not be a new file

bcho892: Also try a story where each text element is a different length

bcho892 (approved):

PR #277, opened 2024-04-19
157 reusable table component

ganznz:


client/src/components/generic/ReusableTable/Table.tsx
bcho892, line 1: I'd recommend using generics so even though we can pass any type in it has to be the same, something like this snippet:
```tsx
interface MyComponentProps<T> {
  data: T[];
}
// notice the trailing comma after <T
const MyComponent = <T,>({ data }: MyComponentProps<T>) => {
  return (
  <div>{JSX content}</div>;
  )
};
export default MyComponent;
```
https://www.freecodecamp.org/news/typescript-generics-with-functional-react-components/

client/src/components/generic/ReusableTable/Table.story.tsx
bcho892, line 1: 👍

bcho892 (changes requested):
Just a small change to make it more robust

bcho892 (approved):

PR #278, opened 2024-04-19
152 pricing logic service

AzizPatel786:


server/src/data-layer/services/PricingService.ts
```diff
@@ -0,0 +1,18 @@
+import { UserAdditionalInfo } from "data-layer/models/firebase"
+import { MembershipType } from "./membershipType"
+
+export default class PricingService {
+  public async getMembershipType(
+    user: UserAdditionalInfo
+  ): Promise<MembershipType> {
```
bcho892, line 7: this this does not need to be async - we are not making any network requests

server/src/data-layer/services/membershipType.ts
bcho892, line 1: 👍

server/src/data-layer/services/PricingService.test.ts
bcho892, line 1: 👍

server/src/data-layer/services/membershipType.ts
bcho892, line 1: nit: file name should be in `PascalCase`

server/src/data-layer/services/PricingService.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (user.university && user.returning) {
+      return MembershipType.UoaReturning
```
bcho892, line 9: For these we should also check if studentid is defined or not:
image

bcho892 (changes requested):
Just some small changes

bcho892 (approved):

PR #281, opened 2024-04-23
Have endpoint to send client secret to user

### LINKED ISSUE #189: [BACKEND] Set up endpoint to generate a checkout section for a user to pay for a membership
**Is your feature request related to a problem? Please describe.**
We will need to use the pricing logic in #152 to find the correct stripe product for the user making the request using #124. We will get the checkout using the functionality implemented in #127 

We should also return the product information using #124 

this should be scoped to the user itself (extract the user data from their JWT) and fail if the user is already a member (think of other edge cases 😄).  Edit the `StripeService` function to allow the checkout session to take the user's `stripe_customer_id`

We should make sure the `return_url` includes a `CHECKOUT_SESSION_ID` like: (`return_url: `${YOUR_DOMAIN}/return?session_id={CHECKOUT_SESSION_ID}``)

**note** we should secure lodge bookings i.e check the specific roles on the user to make sure that only members can try purchasing, likewise, we should make sure only non-members can attempt to purchase club membership.

**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] [Schemas annotated](https://tsoa-community.github.io/docs/annotations.html) if adding new models
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

bcho892:
Closes #189 

## Context

Have set up an endpoint that generates the required client secret and returns ready for use with embedded stripe checkout. This creates a checkout session that expires after ~30 minutes.

Have also switched from using the customer email to making use of the `customer` object in stripe (updated relevant tests) and added the `stripe_id` to the users' data (only usable by the server)

## Screenshots

membership endpoint

image
We get a conflict if users paid recently to allow some leeway for the webhook to do its job (to stop users from double paying)

## Required Stripe configuration

Note that in stripe:
image
The metadata should be set up with the keys/values defined in `server\src\business-layer\utils\StripeProductMetadata.ts`. A wiki page has been created detailing the process: https://github.com/UoaWDCC/uasc-web/wiki/Stripe-Product-Setup


8cw:
Will give this a review tomorrow!

server/src/business-layer/services/StripeService.ts
```diff
@@ -22,6 +36,87 @@ export default class StripeService {
...
+   * @warning This assumes that users won't be paying for membership right after booking
+   *
+   * @param customerId `stripe_id` from the firebase document
+   */
+  public async hasProcessingPaymentIntent(customerId: string) {
+    const { data } = await stripe.paymentIntents.list({
+      customer: customerId
+    })
+    const hasProcessingPaymentIntent = !!data.find(
+      (intent) => intent.status === "processing"
```
8cw, line 64: I notice that the docs for [intent statuses](https://docs.stripe.com/payments/paymentintents/lifecycle#intent-statuses) mention that
> These types of payment methods can take up to a few days to process.

I wonder if it's even worth having this check? Probably won't harm us, but we might want to measure later on how often people hit a processing intent and decide if we can cut this out to simplify the code.

server/src/business-layer/services/StripeService.ts
```diff
@@ -22,6 +36,87 @@ export default class StripeService {
...
+  /**
+   *
+   * Checks for processing payments from a user. Used to avoid
+   * users from double paying after completing a checkout session for membership payments.
+   *
+   * @param customerId `stripe_id` from the firebase document
+   * @param createdMinutesAgo how long ago to check for checkout sessions
+   * @returns true if user has completed checkout session from a default `2` minutes ago
+   */
+  public async hasRecentlyCompletedCheckoutSession(
```
8cw, line 79: I understand the intent of this function (we are attempting to make sure that Stripe posts to our webhook before a user attempts to pay for membership again), but I'm unsure if this is necessary.

The docs for [webhooks on `PaymentIntent`](https://docs.stripe.com/payments/payment-intents/verifying-status#webhooks) makes me think that Stripe provides some level of guarantee that the API won't return a `complete` session until _after_ the webhook has been sent.

We probably need to explore the guarantees of the webhooks from Stripe (e.g., what happens if our backend goes down and a user completes a session? How will Stripe re-deliver the webhook event) to investigate if this function is needed.

server/src/business-layer/services/StripeService.ts
```diff
@@ -35,25 +130,39 @@ export default class StripeService {
...
     line_item: {
       price: string
       quantity: number
     },
-    metadata: Record<string, string>
+    metadata: Record<string, string>,
+    customer_id: string,
+    expires_after_mins: number = 31
   ) {
     const session = await stripe.checkout.sessions.create({
```
8cw, line 154: The docs for [`sessions.create`](https://docs.stripe.com/api/checkout/sessions/object#checkout_session_object-customer_email) mention that for the `customer_email`
> If provided, this value will be used when the Customer object is created. If not provided, customers will be asked to enter their email address. Use this parameter to prefill customer data if you already have an email on file. To access information about the customer once the payment flow is complete, use the customer attribute.

We're currently not providing the `customer_email` here. It's not a problem I think, but might mean that the session email might be different than the firebase email. This probably can end up being the case anyway (e.g., if a user has a different billing email to the UASC email) -- just something to keep in the back of the mind.

server/src/service-layer/controllers/PaymentController.ts
```diff
@@ -0,0 +1,139 @@
...
+         */
+        const activeSession = await stripeService.getActiveSessionForUser(
+          stripeCustomerId,
+          CheckoutTypeValues.MEMBERSHIP
+        )
+        if (activeSession) {
+          const { client_secret, metadata } = activeSession
+          this.setStatus(200)
+          return {
+            clientSecret: client_secret,
```
8cw, line 64: We might want to rename this `clientStripeSecret` or something to indicate it is the stripe secret, maybe.

server/src/service-layer/controllers/PaymentController.ts
```diff
@@ -0,0 +1,139 @@
...
+      // We assume there will only be one active product at a time for memberships (needs to be communicated with admins)
+      const requiredMembershipProduct = requiredMembershipProducts.find(
+        (product) => product.active
+      )
+
+      if (requiredMembershipProduct === undefined) {
+        throw new Error("Product not found")
+      }
+
+      // Note this will throw error if undefined (product wasn't found)
```
8cw, line 113: probably can be removed since the above check will throw the error, not this next line. obviously a nit though.

server/src/service-layer/controllers/PaymentController.ts
```diff
@@ -0,0 +1,139 @@
...
+      const { default_price } = requiredMembershipProduct
+
+      const clientSecret = await stripeService.createCheckoutSession(
+        uid,
+        // Note if the url changes workflows need to be updated to have the deployments work correctly
+        `${process.env.FRONTEND_URL}/register/confirm?session_id={CHECKOUT_SESSION_ID}`,
```
8cw, line 119: Thought this was wrong for a sec, but nope, that's what [stripe wants us to do](https://docs.stripe.com/payments/checkout/custom-success-page)!

8cw (approved):
Looks great to me, well done with the tests as well.

This will serve well for the booking system -- we might have to refactor some of the functions to be more extensible there, but we'll get there when we get there.

server/src/business-layer/services/StripeService.ts
bcho892: Stripe seems to retry the webhook for up to 3 days: https://docs.stripe.com/webhooks#retries

bcho892: I was thinking its more to account for how we have to also refresh user tokens on the front end (to check if their account has been updated) so they might have a window to accidentally make another call - but it's probably true their first instinct will not to be pay again.

server/src/business-layer/services/StripeService.ts
bcho892: Have created #287

server/src/business-layer/services/StripeService.ts
bcho892: If we remove this the claims check in this endpoint would be the only safety net against this

8cw: > I was thinking its more to account for how we have to also refresh user tokens on the front end (to check if their account has been updated) so they might have a window to accidentally make another call - but it's probably true their (user's) first instinct will not to be pay again.

With this current implementation, I don't think this will be allowed by the server if the webhook is processed at the correct time (as you say, the claims check in the endpoint).

Our server uptime should be pretty good and probably would give sufficient confidence to not need to do this check (and instead rely on Stripe delivering the webhook event between server crashes) -- i.e., the server will correctly reject the request.

I'd say we leave this for now, as it's not going to do any harm (it's only 1 extra API call to Stripe).

server/src/business-layer/services/StripeService.ts
bcho892: Actually good point raised here - in the `createNewUser()` for stripe service we force them to use their firebase email but that's probably not the best choice. I'll remove that from the method.

server/src/business-layer/services/StripeService.ts
bcho892: one thing to note is WDCC have deployed our app on a serverless platform, idk how that'll affect it

PR #294, opened 2024-04-25
Created generic alert components

Kartik-M24:


client/src/components/generic/AlertsComponent/AlertsComponent.tsx
```diff
@@ -0,0 +1,122 @@
...
+  message: string
+  variant?: "success" | "notification" | "error"
+}
+
+const AlertsComponent = ({ variant, message }: AlertsInputProp) => {
+  if (variant === "success") {
+    return (
+      <button className="flex h-12 w-full items-center justify-between rounded-lg bg-[#109D27] text-left uppercase text-white">
+        <div className="ml-10 mr-5">
+          <svg
```
bcho892, line 13: The svgs should already be in `client/src/assets`, and you can import using the syntax like: 

```ts
import LeftArrowIcon from "assets/icons/leftarrow.svg?react"
// then
 <LeftArrowIcon className="fill-white" />
```

client/src/components/generic/AlertsComponent/AlertsComponent.tsx
```diff
@@ -0,0 +1,122 @@
...
+            className="mr-10 h-6 w-6"
+          >
+            <path
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              d="M6 18 18 6M6 6l12 12"
+            />
+          </svg>
+        </div>
+      </button>
```
bcho892, line 117: note: it is semantically *incorrect* and an anti-pattern to put content all inside a button - buttons should only be used to contain content that should be clicked.

Kartik-M24:
Screenshot 2024-04-25 224412

Kartik-M24:
image

Kartik-M24:
image

client/src/components/generic/AlertsComponent/AlertsComponent.tsx
```diff
@@ -0,0 +1,51 @@
...
+    )
+  } else {
+    return (
+      <div className="flex h-12 w-full items-center justify-between rounded-lg bg-[#FF6D04] text-left uppercase text-white">
+        <div className="ml-10 mr-5 h-5 w-5">
+          <BellIcon className="fill-white" />
+        </div>
+        <span className="absolute left-20 ml-10"> {message} </span>
+        <div className="mr-10 h-5 w-5">
+          <XIcon className="fill-white" />
```
bcho892, line 44: since we have a close button, we should allow consumers to add a onclick handler (or `onClose` in this case). We should also give an option to hide the close button, think about how you can do this with props.

bcho892:
> image

Looks good ~ have you tested how it looks when you make the screen smaller?

PR #297, opened 2024-04-26
[DRAFT] 188 signup endpoint

jeffplays2005:


server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    let user: UserRecord
+    console.log(requestBody) // debug
+    try {
+      // create user in auth service
+      // const user = await authService.createUser(requestBody.email)
+      user = await authService.createUser(requestBody.email)
+      // create user in firestore
+      await userService.createUserData(user.uid, requestBody.user)
+    } catch (e) {
```
bcho892, line 25: You might want to add a message before the console so you can debug

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,52 @@
...
+    // create services
+    const userService = new UserDataService()
+    const authService = new AuthService()
+
+    let user: UserRecord
+    console.log(requestBody) // debug
+    try {
+      // create user in auth service
+      // const user = await authService.createUser(requestBody.email)
+      user = await authService.createUser(requestBody.email)
```
bcho892, line 22: Might want to destructure here
```suggestion
      { uid } = await authService.createUser(requestBody.email)
```

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,52 @@
...
+import UserDataService from "data-layer/services/UserDataService"
+import { UserRecord } from "firebase-admin/auth"
+import { UserSignupBody } from "service-layer/request-models/UserSignupRequests"
+import { Body, Controller, Post, Route, SuccessResponse } from "tsoa"
+
+@Route("signup")
+export class UserSignup extends Controller {
+  @Post()
+  @SuccessResponse(200)
+  // return a JWT token at the end
```
bcho892, line 11: Ok ignore my comment on removing jwt, this is right

server/src/service-layer/request-models/UserSignupRequests.ts
```diff
@@ -0,0 +1,7 @@
+import { UserAdditionalInfo } from "data-layer/models/firebase"
+import { CommonResponse } from "service-layer/response-models/CommonResponse"
+
+export interface UserSignupBody extends CommonResponse {
```
bcho892, line 4: I meant to extend the response for the enpoint so the format would be:

```ts
{
jwtToken: string,
// from commonresponse
message: optional,
error:optional
}
```

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,37 @@
...
+export class UserSignup extends Controller {
+  @Post()
+  @SuccessResponse(200, "Signup successful")
+  // return a JWT token at the end
+  public async signup(@Body() requestBody: UserSignupBody): Promise<string> {
+    const userService = new UserDataService()
+    const authService = new AuthService()
+
+    // Omit membership to avoid users from creating an account with admin
+    const userInfo: Omit<UserAdditionalInfo, "membership"> = requestBody.user
```
bcho892, line 17: we should omit this information in the request body type, so that it doesn't appear on the our API docs. (Also need to omit `stripe_id`

server/src/data-layer/adapters/FirestoreUtils.ts
```diff
@@ -5,5 +5,10 @@ export const firestoreTimestampToDate = (firestoreDate: Timestamp) => {
 }
 
 export const dateToFirestoreTimeStamp = (date: Date) => {
-  return Timestamp.fromDate(date)
+  let output = Timestamp.fromDate(date)
```
bcho892, line 8: 👍 could probably write a single unit test with `new Date(0)`

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,45 @@
...
+        userRecord.uid,
+        undefined
+      )
+      this.setStatus(200)
+      return { jwtToken, uid: userRecord.uid }
+    } catch (error) {
+      console.error(error)
+      this.setStatus(500) // server error
+      return {
+        jwtToken: null,
```
bcho892, line 40: you can avoid this by making jwtToken optional

server/src/data-layer/adapters/FirestoreUtils.test.ts
bcho892, line 1: 👍

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,45 @@
...
+  ): Promise<UserSignupResponse> {
+    const userService = new UserDataService()
+    const authService = new AuthService()
+
+    // Received userInfo omits membership and stripe_id
+    const userInfo = requestBody.user
+
+    try {
+      // Create user data in Auth Service
+      const userRecord = await authService.createUser(requestBody.email)
```
bcho892, line 23: nit: i would still have a separate try catch so we can return `{error: email in use}` etc

bcho892 (approved):

server/src/service-layer/controllers/UserSignup.ts
jeffplays2005: Ah yes, I remember thinking about that conflict before but somehow deleted it some while ago. Ty! Will implement.

bcho892: actually ignore that, in the 500 response you can just make error = the caught error

server/src/data-layer/adapters/FirestoreUtils.test.ts
```diff
@@ -0,0 +1,11 @@
+import { dateToFirestoreTimeStamp } from "./FirestoreUtils"
+
+describe("FirestoreUtils Integration test", () => {
```
bcho892, line 3: nit: this is a unit test

server/src/service-layer/controllers/UserSignup.ts
jeffplays2005: Oops, no worries, just created another try/catch for the different 409 conflict.

server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -0,0 +1,49 @@
...
+      const jwtToken = await authService.createCustomToken(
+        userRecord.uid,
+        undefined
+      )
+      this.setStatus(200)
+      return { jwtToken, uid: userRecord.uid }
+    } catch (error) {
+      console.error(error)
+      this.setStatus(500) // server error
+      return {}
```
bcho892, line 46: nit: id return the error here

PR #314, opened 2024-05-01
126 stripe service prod update

AzizPatel786:


server/src/business-layer/services/StripeService.test.ts
```diff
@@ -1,5 +1,6 @@
 import StripeService from "./StripeService"
 import { productMock } from "test-config/mocks/Stripe.mock"
+import Stripe from "stripe"
```
bcho892, line 3: Nit: Don't think we need to import this?

bcho892 (approved):
Just small nit in tests, but lgtm

server/src/business-layer/services/StripeService.test.ts
AzizPatel786: I was getting an error because the type of `param` wasn't getting detected in line 28 hence, I had to explicitly mention it's an object type of `Stripe.ProductUpdateParams`

bcho892: Ok makes sense

Issue #237, opened 2024-04-07
[BACKEND] Allow Stripe Service to retrieve checkout session based on payment intent


bcho892:
**Is your feature request related to a problem? Please describe.**
Blocks #133 #132 
https://docs.stripe.com/api/checkout/sessions/list#list_checkout_sessions-payment_intent
**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] [Schemas annotated](https://tsoa-community.github.io/docs/annotations.html) if adding new models
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #315, opened 2024-05-01
132 - checkout session logic

jeffplays2005:
Draft pull request
Updated to include the correct checkout session type. Also updated to fetch the session type based on the Stripe api checkout sessions fetch.

### DIFF
--- a/.github/workflows/deploy-server.production.yml
+++ b/.github/workflows/deploy-server.production.yml
@@ -12,6 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: superfly/flyctl-actions/setup-flyctl@master
-      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
+      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_WEBHOOK_SECRET=${{ secrets.STAGING_SERVER_STRIPE_WEBHOOK_SECRET }} STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
       - run: flyctl secrets deploy --config fly.production.toml
       - run: flyctl deploy --remote-only --config fly.production.toml
--- a/.github/workflows/deploy-server.production.yml
+++ b/.github/workflows/deploy-server.production.yml
@@ -12,6 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: superfly/flyctl-actions/setup-flyctl@master
-      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_WEBHOOK_SECRET=${{ secrets.STAGING_SERVER_STRIPE_WEBHOOK_SECRET }} STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
+      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_WEBHOOK_SECRET=${{ secrets.PROD_SERVER_STRIPE_WEBHOOK_SECRET }} STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
       - run: flyctl secrets deploy --config fly.production.toml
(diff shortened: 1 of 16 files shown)


jeffplays2005:
Need to create route tests(need to be completed locally and documented after).

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -1,38 +1,66 @@
...
+    if (event.type === "payment_intent.succeeded") {
+      // console.log("[WEBHOOK] received: payment_intent.succeeded")
+      console.log(request.body)
+      const userService = new UserDataService()
+      const authService = new AuthService()
+      // Stripe PaymentIntent
+      const { id } = event.data.object
+      // Fetch the checkout session from the PaymentIntent ID
+      const stripeSession: Stripe.Checkout.Session =
+        await stripe.checkout.sessions.retrieve(id)
```
bcho892, line 39: i believe we have a method for this in `StripeService` already, which was created in #237

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -1,38 +1,66 @@
...
-
-      return this.setStatus(200) // set status to 200 as success
     } catch (err) {
       console.error(err)
-      // set status to 400 due to bad request
-      return this.setStatus(400)
+      return this.setStatus(401) // unauthorized request
+    }
+    // console.log(event.type)
+    if (event.type === "payment_intent.succeeded") {
```
bcho892, line 30: can we use a switch?

bcho892:
> Need to create route tests(need to be completed locally and documented after).

You can have a look at how I've stubbed stripe in `routes.test.ts`

server/src/service-layer/controllers/StripeWebhook.ts
jeffplays2005: Whoops, yep just made the changes, has been a long time 😂

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -26,40 +27,45 @@ export class StripeWebhook extends Controller {
...
+    switch (event.type) {
+      case "payment_intent.succeeded":
+        if (event.type === "payment_intent.succeeded") {
+          console.debug("[WEBHOOK] received payment_intent.succeeded")
+          // Stripe PaymentIntent
+          const { customer } = event.data.object
+          // Fetch the checkout session from the PaymentIntent ID
+          const stripeSession =
+            await stripeService.retrieveCheckoutSessionFromPaymentIntent(
+              "checkout.session",
```
bcho892, line 44: This does not seem right, we want to pass a `payment_intent` id

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -26,40 +27,45 @@ export class StripeWebhook extends Controller {
...
+              customer.toString(),
```
bcho892, line 45: we can omit this, because there will only ever be one `checkout_session` for each `payment_intent`

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -26,40 +27,45 @@ export class StripeWebhook extends Controller {
...
+              "complete"
```
bcho892, line 46: not necessarily true because the payment intent could be fired before

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -26,40 +27,45 @@ export class StripeWebhook extends Controller {
...
-          console.log(
-            "successfully received checkout.session.completed and added membership"
+    switch (event.type) {
+      case "payment_intent.succeeded":
+        if (event.type === "payment_intent.succeeded") {
+          console.debug("[WEBHOOK] received payment_intent.succeeded")
+          // Stripe PaymentIntent
+          const { customer } = event.data.object
+          // Fetch the checkout session from the PaymentIntent ID
+          const stripeSession =
```
bcho892, line 42:
```suggestion
          const [stripeSession] =
```

this will automatically give you the first element of the array (the array size is guaranteed to be at max 1

bcho892 (changes requested):
We should have a discussion some time on how to manually test that this works with the other integrations, but also I can spot some small logical errors

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -1,38 +1,71 @@
...
+    // Ensure security of the endpoint by constructing an event
+    let event: Stripe.Event
     try {
-      const event: Stripe.Event = stripe.webhooks.constructEvent(
+      event = stripe.webhooks.constructEvent(
         request.rawBody,
         request.headers["stripe-signature"],
-        // process.env.STRIPE_LOCAL // test local api secret
-        process.env.STRIPE_API_SECRET
+        process.env.STRIPE_LOCAL
```
bcho892, line 23: can we update the workflows if we are going to change the name of env variables

jeffplays2005: whoops sorry forgot to change that, should be STRIPE_API_SECRET, the LOCAL one is for local testing.

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -1,38 +1,71 @@
...
+          )
+            return this.setStatus(400) // bad request, not the memberhip we want
+          try {
+            // need to update firestore
+            await userService.editUserData(uid, {
+              membership: "member" // only update their membership
+            })
+            // need to add member claim to user
+            await authService.setCustomUserClaim(uid, "member")
+            console.debug("[WEBHOOK] added membership")
```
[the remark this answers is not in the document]
jeffplays2005: Added the UID to the debug line!

.github/workflows/deploy-server.production.yml
```diff
@@ -12,6 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: superfly/flyctl-actions/setup-flyctl@master
-      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
+      - run: flyctl secrets set --config fly.production.toml "GOOGLE_SERVICE_ACCOUNT_JSON=$(echo ${{ secrets.PROD_SERVER_SERVICE_ACCOUNT }} | base64 --decode)" STRIPE_WEBHOOK_SECRET=${{ secrets.STAGING_SERVER_STRIPE_WEBHOOK_SECRET }} STRIPE_API_SECRET=${{ secrets.PROD_SERVER_STRIPE_API_SECRET }} STRIPE_API_KEY=${{ secrets.PROD_SERVER_STRIPE_API_KEY  }} FRONTEND_URL=https://uasc-dev.wdcc.co.nz/
```
bcho892, line 15: The secret name should be different from staging

bcho892 (approved):
GJ!

PR #323, opened 2024-05-02
Created CRUD for BookingSlotsService and tests for them.

Kartik-M24:


Kartik-M24:
I don't think I've correctly implemented getBookingSlotByDate so wanted to get someone to clarify that for me. Thanks in advance :)

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -0,0 +1,43 @@
...
+  public async getBookingStripeProductId(productId: string) {
+    const result = await FirestoreCollections.bookingSlots
+      .where("stripe_product_id", "==", productId)
+      .get()
+    const stripeProductIdArray = result.docs.map((docs) => docs.data())
+    return stripeProductIdArray
+  }
+
+  public async getBookingSlotByDate(date: Timestamp) {
+    // Not sure if this one works am I supposed to read between start and end date?
```
Kartik-M24, line 21: This is the part I'm specifically confused about.

8cw (approved):
LGTM, a small nit on the function name for one of them.

I haven't read the tests, so will make sure they are sound tomorrow if I get time. But feel free to merge otherwise.

server/src/data-layer/services/BookingSlotsService.ts
Kartik-M24: AHH okay that makes more sense I didn't know this, thanks for lmk.

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Timestamp } from "firebase-admin/firestore"
+
+export default class BookingSlotService {
+  // Create
+  public async createBookingSlot(bookingSlotData: BookingSlot) {
+    return await FirestoreCollections.bookingSlots.add(bookingSlotData)
+  }
+
+  // Read
+  public async getBookingStripeProductId(productId: string) {
```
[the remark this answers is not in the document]
Kartik-M24: Okay noted, thanks.

server/src/data-layer/adapters/FirestoreCollections.ts
```diff
@@ -24,7 +28,8 @@ const firestore = Object.assign(
 
 const db = {
   users: firestore.collection<UserAdditionalInfo>("users"),
-  bookings: firestore.collection<Booking>("bookings")
+  bookings: firestore.collection<Booking>("bookings"),
+  bookingSlots: firestore.collection<BookingSlot>("bookingSlots")
```
bcho892, line 32: I think the collection is called `booking_slots` instead of `bookingSlots`.

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -0,0 +1,43 @@
+import FirestoreCollections from "data-layer/adapters/FirestoreCollections"
+import { BookingSlot } from "data-layer/models/firebase"
+import { Timestamp } from "firebase-admin/firestore"
+
+export default class BookingSlotService {
+  // Create
+  public async createBookingSlot(bookingSlotData: BookingSlot) {
+    return await FirestoreCollections.bookingSlots.add(bookingSlotData)
```
bcho892, line 8: your tests are currently failing because when you create a document this way it gets assigned a random id, while you try to get it with `TEST_SPID`.

To fix this you have 2 choices:

```ts
  public async createBookingSlot(
    stripeId: string,
    bookingSlotData: BookingSlot
  ) {
    return await FirestoreCollections.bookingSlots
      .doc(stripeId)
      .set(bookingSlotData)
  }
```
Set the id in this function

**OR**

use the return value of this function in the tests

```ts
    const { id } = await bookingSlotService.createBookingSlot(bookingSlotData)
```

bcho892: actually nvm you seem to be getting it by stripe id anyway

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  beforeEach(() => {
+    bookingSlotService = new BookingSlotsService()
+  })
+
+  it("Should create a booking slot", async () => {
+    await bookingSlotService.createBookingSlot(bookingSlotData)
+    const data = await bookingSlotService.getBookingStripeProductId(TEST_SPID)
+
+    expect(data).not.toBe(undefined)
```
bcho892, line 35: keep in mind this method returns an array so it will fail here

server/src/data-layer/services/BookingSlotsService.ts
bcho892: image

If you're talking about delete booking slot it passes because 
```ts
const data = await bookingSlotService.getBookingStripeProductId(TEST_SPID)
```
returns undefined no matter what rn

bcho892: so yea you should choose one of the options, TLDR #1 will mean you can keep using `TEST_SPID` and option #2 means you'll have to update most your tests

Kartik-M24: Okay thanks Benson.

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  // Read
+  public async getBookingStripeProductId(productId: string) {
+    const result = await FirestoreCollections.bookingSlots
+      .where("stripe_product_id", "==", productId)
```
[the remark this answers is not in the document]
Kartik-M24: Okay I think I've implemented most of the changes you've suggested, the update method still fails the test, could you help me with that?

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      ...bookingSlotData
+    })
+  })
+
+  it("Should update booking slot", async () => {
+    const { id } = await new BookingSlotsService().createBookingSlot(
+      bookingSlotData
+    )
+    await new BookingSlotsService().updateBookingSlot(id, { max_bookings: 10 })
+    const doc = await FirestoreCollections.bookings.doc(id).get()
```
bcho892, line 61: I think you are using the wrong collection

server/src/data-layer/services/BookingSlotsService.ts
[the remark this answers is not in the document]
bcho892: See my comment https://github.com/UoaWDCC/uasc-web/pull/323#discussion_r1588594577

server/src/data-layer/services/BookingSlotsService.test.ts
Kartik-M24: Yup that was it I'm and idiot thanks Benson.

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -3,13 +3,14 @@ import { BookingSlot } from "data-layer/models/firebase"
 import { Timestamp } from "firebase-admin/firestore"
 
 export default class BookingSlotService {
+  static createBookingSlot: any
```
8cw, line 6: Is this meant to be here?

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  it("Should create a booking slot", async () => {
+    const { id } = await new BookingSlotsService().createBookingSlot(
+      bookingSlotData
+    )
+    const doc = await FirestoreCollections.bookingSlots.doc(id).get()
+    const data = doc.data()
+
+    expect(data).not.toBe(undefined)
+    expect(data).toEqual({ ...bookingSlotData }) // not sure if this is correct
```
8cw, line 28:
```suggestion
    expect(data).toEqual(bookingSlotData)
```

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    await new BookingSlotsService().createBookingSlot(bookingSlotData)
+    const bookingSlotsByStripeProductID =
+      await new BookingSlotsService().getBookingSlotsStripeProductId(
+        "product_id_value"
+      )
+
+    expect(bookingSlotsByStripeProductID).not.toBe(undefined)
+    expect(bookingSlotsByStripeProductID[0]).toEqual({
+      ...bookingSlotData // not sure if this is correct
+    })
```
8cw, line 41:
```suggestion
    expect(bookingSlotsByStripeProductID[0]).toEqual(bookingSlotData)
```
It's okay if our test functions throw errors (the Jest harness will take care of it), so we can remove these `.not.toBe(undefined)` checks.

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -0,0 +1,44 @@
...
+  public async getBookingSlotsStripeProductId(productId: string) {
+    const result = await FirestoreCollections.bookingSlots
+      .where("stripe_product_id", "==", productId)
+      .get()
+    const stripeProductIdArray = result.docs.map((docs) => docs.data())
+    return stripeProductIdArray
+  }
+
+  public async getBookingSlotByDate(date: Timestamp) {
+    // Not sure if this one works am I supposed to read between start and end date?
```
8cw, line 22:
```suggestion
```

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      bookingSlotData
+    )
+    const doc = await FirestoreCollections.bookingSlots.doc(id).get()
+    const data = doc.data()
+
+    expect(data).not.toBe(undefined)
+    expect(data).toEqual({ ...bookingSlotData }) // not sure if this is correct
+  })
+
+  it("Should get booking slots based on stripe product id", async () => {
```
8cw, line 31: I would love for us to better test the query here to make sure it works. Here's a scenario I think works well for us:
1. insert `bookingSlotData`
2. insert a random booking slot (that has different data)
3. insert `bookingSlotData` (except the date is the next date)
we could create these three objects before our test handlers (we currently only create one, `bookingSlotData`).

this way the query will test both
- the fact that it will not select the random booking slot
- the fact that it will select _all_ booking slots that have the same Stripe ID.

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  it("Should get booking slots based on date", async () => {
+    await new BookingSlotsService().createBookingSlot(bookingSlotData)
+    const bookingSlotsByDate =
+      await new BookingSlotsService().getBookingSlotByDate(timestamp)
+
+    expect(bookingSlotsByDate).not.toBe(undefined)
+    expect(bookingSlotsByDate.length).toBe(1)
+    expect(bookingSlotsByDate[0]).toEqual({
+      ...bookingSlotData
+    })
```
8cw, line 53:
```suggestion
    expect(bookingSlotsByDate.length).toBe(1)
    expect(bookingSlotsByDate[0]).toEqual(bookingSlotData)
```

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+        "product_id_value"
+      )
+
+    expect(bookingSlotsByStripeProductID).not.toBe(undefined)
+    expect(bookingSlotsByStripeProductID[0]).toEqual({
+      ...bookingSlotData // not sure if this is correct
+    })
+  })
+
+  it("Should get booking slots based on date", async () => {
```
8cw, line 44: Similar to above, if we insert two booking slots with different dates this would be a good test.

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      bookingSlotData
+    )
+    await new BookingSlotsService().updateBookingSlot(id, { max_bookings: 10 })
+    const doc = await FirestoreCollections.bookingSlots.doc(id).get()
+    const data = doc.data()
+
+    expect(data).toEqual({
+      ...bookingSlotData,
+      max_bookings: 10
+    })
```
8cw, line 67: Awesome!

8cw (changes requested):
Just had a quick run through the test functions. I think we can improve some of the scenarios we are testing to be a little bit more realistic and rigorous in the way it tests, but this is pretty close to being merge-able, so good stuff.

server/src/data-layer/services/BookingSlotsService.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    await new BookingSlotsService().createBookingSlot(bookingSlotData)
+    const bookingSlotsByDate =
+      await new BookingSlotsService().getBookingSlotByDate(timestamp)
+
+    expect(bookingSlotsByDate).not.toBe(undefined)
+    expect(bookingSlotsByDate.length).toBe(1)
+    expect(bookingSlotsByDate[0]).toEqual(bookingSlotData)
+  })
+
+  it("Should get booking slots between date range", async () => {
```
bcho892, line 73: nit: should probably name these more descriptively

bcho892 (approved):

8cw (approved):
LGTM

PR #326, opened 2024-05-03
321-Remove membership references

jeffplays2005:
Removed membership from Firebase model.

Removed fire store updates in the following endpoints:

* StripeWebhook
  * Removed firestore UserService updates.
  * Updated route testing.
* UserController
  * Changed demote/promote that use UserService.
  * Removed corresponding tests that test for the membership field.
* UserSignup
  * Remove the omitting of membership from the UserSignupRequests interface.
  * Removed setting membership to guest while creating user data.
  * Removed corresponding tests that test for specific membership type.
* PaymentController
  * Update route testing to remove membership types.
  
Services updated:

* UserDataService
  * Update getFilteredUsers to remove filtering on membership type.
  * Remove corresponding tests based on membership types.
  * Update tests to use the generalised userInfoMocks
  * Removed omitting in EditSelfRequestBody interface.
  
Mocks updated:

* Removed guest and admin mocks.
* Changed admin mock to a second "general" userInfoMock
  * Changed PricingService to correct mock names.
* Updated and neatened route mocks to reflect the removal of membership.


server/src/service-layer/controllers/UserSignup.ts
```diff
@@ -29,10 +29,7 @@ export class UserSignup extends Controller {
 
     try {
       // Create document with user info
```
bcho892, line 31: nit: do we still need this comment

bcho892 (approved):

PR #335, opened 2024-05-06
Implement mobile navbar

bcho892:
Have implemented the new design for mobile, made sure desktop doesn't break. Shadow colour also updated to reflect new design. 

This was done fully with media queries so there should be no performance hit that would occur from js-styling.

Screenshot_20240506-202937

image
Most compact version

image
Normal mobile

image
Desktop


asun555 (approved):
LGTM

PR #340, opened 2024-05-09
Created success screen after user has made a booking

Kartik-M24:


Kartik-M24:
image
Small mobile view

Kartik-M24:
image

Kartik-M24:
I haven't grabbed the dates as I was a bit unsure where I would import them from, is it from the data-layer?
I also wasn't sure what the bookings URL was for the button

bcho892:
image
Padding vertically seems a bit off, but gj on the mobile view! (IMO the container should have a smaller horizontal padding on smaller screens so it fills more). 

For the dates we'd likely make use of search params https://reactrouter.com/en/main/hooks/use-search-params to grab the dates (we will tell stripe to redirect to a page with those as the params)

bcho892:
> I haven't grabbed the dates as I was a bit unsure where I would import them from, is it from the data-layer? I also wasn't sure what the bookings URL was for the button

Have a look at `routes.tsx`

client/src/components/composite/SuccessfulBooking/SuccessfulBooking.tsx
```diff
@@ -0,0 +1,37 @@
...
+  endDate: string
+}
+
+const SuccessfulBooking = ({ startDate, endDate }: SuccessfulBookingProps) => {
+  return (
+    <FullPageBookLodgeImage>
+      <div className="text-dark-blue-100 mb-auto mr-auto px-2 py-6 align-text-top italic md:ml-32 md:px-20 md:py-20">
+        <h2>Book the Lodge</h2>
+      </div>
+      <div className=" xs:-top-72 relative mx-auto mb-48 h-auto w-10/12 rounded-sm border border-black bg-white px-2 py-2 md:w-6/12 lg:-mb-10">
```
bcho892, line 14: I'd suggest having this full width to align with the heading

client/src/pages/BookingSuccessComponent.tsx
```diff
@@ -0,0 +1,11 @@
+import SuccessfulBooking from "components/composite/SuccessfulBooking/SuccessfulBooking"
+import { useSearchParams } from "react-router-dom"
+
+const bookingSuccess = () => {
```
bcho892, line 4: React components must be in `PascalCase`

client/src/components/composite/SuccessfulBooking/SuccessfulBooking.tsx
```diff
@@ -0,0 +1,37 @@
+import FullPageBookLodgeImage from "components/generic/FullPageBookLodgeImage/FullPageBookLodgeImage"
+
+type SuccessfulBookingProps = {
+  startDate: string | null
```
bcho892, line 4: prefer the optional tag, as that handles both null and undefined
```suggestion
  startDate?: string
```

client/src/pages/BookingSuccessComponent.tsx
bcho892, line 1: nit: Probably could just call this file `BookingSuccess`

bcho892 (changes requested):
Nice job on creating the components, now we should make this a subroute of the `bookings` route - have a look in the `Login` page to see how it was done with the password reset form, and refer to docs for [react router](https://www.robinwieruch.de/react-router-nested-routes/)

Kartik-M24:
Screenshot 2024-05-20 234905

bcho892:
there seems to be a bit of overflow on the page, we shouldn't have this on desktop

bcho892 (approved):
nit: next time branch off master so we don't get the garbage commits, you can technically fix this with cherry picking but it's better to avoid that (it's fine for this one)

PR #343, opened 2024-05-10
Refactor user schema

bcho892:
We had an issue where `confirmEmail` was being part of the payload for the `signup` endpoint, causing validation to fail.


eddiew16597 (approved):
approved

asun555 (approved):
LGTM

PR #356, opened 2024-05-15
Update membership types, make membership selectable on frontend

bcho892:
The previous implementation included the wrong types of membership these have been replaced with:

- "uoa_student",
- "non_uoa_student",
- "returning_member",
- "new_non_student"

Have also allowed selection of the membership type on the front end, as well as fixed the issue with infinite retries leading to it always displaying loading on error
(yea ik it says `backend`)

payment sel

image


asun555 (approved):
LGTM

PR #360, opened 2024-05-19
336 profile page

zlrkw11:
closes #336

PR #317, opened 2024-05-01
303 success section

zlrkw11:
closes #303

PR #366, opened 2024-05-21
Create presentation for bookings page (ignoring ToC & tabs)

bcho892:
## Interactive Calendar
calendar
(Note styling is outdated in the above, calendar no longer has the weird alignment)

The updated styling is: 
image

Also restored #317 which was overwritten in #360. The route is protected and displays a prompt for unpaid/logged out users.

Made the success screen share the same background, and let the calendar page the index route of `/bookings`

image


zlrkw11:
image
image
the calender view will glitch in height when you try to go through the months. Mostly looks good to me.

bcho892:
> image image the information panel will glitch in height when you try to go through the months and I assume its because some months have more dates to display, but the info panel should have no change in appearance. image this part should align with the calender at the top horizontally.

Yeah thats fair thanks for pointing it our

zlrkw11 (approved):

PR #377, opened 2024-05-24
Make workflow fail for frontend if there are TS errors

bcho892:
Right now the frontend still passes builds even if there are type errors, allowing *broken* code to make it into `master`. 

This runs `tsc` before builds so type errors are caught. 

Also fixed the existing errors


bcho892:
yea...
image

asun555 (approved):
LGTM

Issue #149, opened 2024-03-16
[BACKEND] Paginate all users endpoint


bcho892:
**Is your feature request related to a problem? Please describe.**
Use offset based pagination for this.

You will have to fetch all (100 at a time) users 

**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #384, opened 2024-06-02
Create members view for admin page with functionality implemented based on current backend endpoints

bcho892:
# Admin Members View
Leads into all the admin frontend tickets
adminmembers

## Implemented

- Promote/Demote users
- Seaching Users (By name and email)
- Admin link in navbar if user is admin

## Next Steps

- Need to create prompt for admin to create users
- Decide if we want to do filtering "by tags" (My vote is it belongs in *nice to have*)
- Integrate user deletion endpoint
- Create prompt for admin to edit user details
- Make backend endpoint for fetching users include: (this task is included in #149)
   - Membership status
   - Email
   - Date Joined
- Possibly a custom alert message as opposed to the native `window.alert()`

Some additional changes made to improve existing:
- Updated the token scripts to allow removing and adding admin claims (wiki will be updated)
- Made navbar menu close on click outside (desktop view)
- Made admin navbar not do the jank on smaller screens (**even though this should not be a use case**)
- Strapped on a bunch of functionality to the table component so we can:
    - Add custom options to each row (now you *have* to pass in an identifier for the row - generally in the form of a uid)
    - Paginate between each page (there is now an optional param of how many items to show per page)


server/src/business-layer/security/Authentication.ts
```diff
@@ -32,7 +32,10 @@ export function expressAuthentication(
                     "No Scope"
                   )
                 }
-                if (!(scope in user.customClaims)) {
+                if (
+                  !(scope in user.customClaims) ||
```
bcho892, line 36: wasn't treating `member:false` or `admin:false` as invalid

asun555 (approved):
LGTM

Issue #390, opened 2024-06-03
[FRONTEND] Create functional page allowing admin to control date availability


bcho892:
**Is your feature request related to a problem? Please describe.**

We require a view that makes use of #381 to set dates available in an automated way, and also displays the available ones for reference (this will have to use #382).

*May* require redesign later, however having base functionality (calling the endpoints #391 #381) is enough

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

Issue #349, opened 2024-05-12
[BACKEND] Set up endpoint for admin to make date(s) available for booking 


bcho892:
**Is your feature request related to a problem? Please describe.**

The request body should contain:
```ts
{
Date: Timestamp[],
Description: optional,
Title: string
}
```

We want to create a new product (if required) in stripe with the details using #125 and also create a firestore document in the `booking_slots` using #186 (if required). We should NOT allow duplicate stripe products to be created so this should be checked *first*. 

>[!NOTE]
>You may want to first check firestore for a document with the date, if it exists, change the max_slots status to > 0

~~Also remove the `stripe_product_id` from `booking_slot`, replace with `is_active`~~

**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] [Schemas annotated](https://tsoa-community.github.io/docs/annotations.html) if adding new models
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

bcho892:
dont require removing `stripe_product_id` anymore because that's done in #133

PR #391, opened 2024-06-04
Allow admins to make dates unavailable using endpoint

bcho892:
Mostly a copy-paste job of #349, added extra test to make sure it works for skipped dates:
i.e
```
has booking slot / no booking slot / has booking slot
```
Also refactored the names to be plural instead

Is needed for (is blocking) #390


asun555 (approved):
LGTM

Issue #371, opened 2024-05-24
[FRONTEND] Create bookings view for admin page


bcho892:
**Is your feature request related to a problem? Please describe.**

>[!WARNING]
>Deleting bookings *should* be included in this ticket

Need to call endpoint in #415 

![Image](https://github.com/UoaWDCC/uasc-web/assets/100653148/e2659240-b9c6-4827-bbda-16d1b9600cc9)

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #395, opened 2024-06-05
390 frontend create admin availablity page

bcho892:
# Admin availability page

availability

Good reference for the work to be done in #371.

Basically calls the endpoints made in #391 and #381. Probably requires redesign (consider the functionality proposed by #390), but is functionally working. 

Login script is updated so that we also get member claims when requesting an admin login.

Overwrote the *payments* tab as that is not needed.


asun555 (approved):
LGTM

PR #399, opened 2024-06-06
Fix bad input and close navbar on route change

bcho892:
Straightfoward fixes:

- dont handle untruthy values in slot input
- listen for changes in current route (using `useLocation` hook)


asun555 (approved):
LGTM

PR #402, opened 2024-06-07
338 endpoint all user bookings

ganznz:


server/src/service-layer/controllers/BookingController.ts
```diff
@@ -1,12 +1,61 @@
...
+} from "tsoa"
+import { firestoreTimestampToDate } from "data-layer/adapters/DateUtils"
 
 @Route("bookings")
 export class BookingController extends Controller {
+  @SuccessResponse("200", "Found bookings")
+  @Security("jwt", ["member"])
+  @Get()
+  public async getAllBookings(
+    @Query() uid: string
```
bcho892, line 27: the problem with including the UID like this is that members could technically query the bookings of others if they know the `uid`. The safer way to do it is extract their `uid` from the user object returned in the request (which we do in the `Authentication.ts` middleware). 

have a look at the `getSelf()` method in `server/src/service-layer/controllers/UserController.ts`

server/src/middleware/routes.test.ts
```diff
@@ -604,6 +604,82 @@ describe("Endpoints", () => {
...
+      await bookingDataService.createBooking({
+        user_id: MEMBER_USER_UID,
+        booking_slot_id: bookingSlotId1,
+        stripe_payment_id: ""
+      })
+
+      const res = await request
+        .get("/bookings")
+        .set("Authorization", `Bearer ${guestToken}`)
+        .query({ uid: MEMBER_USER_UID })
```
bcho892, line 679: following the other comment, the request wouldn't need these

bcho892 (changes requested):
code is laid out nicely and clean PR 👍. 

Just some issues on security

bcho892 (approved):

PR #406, opened 2024-06-07
Correctly display users ski/snowboard status on profile page

ganznz:


bcho892 (approved):
just a note, should include that it closes #406, in the "development" tab on the right

client/src/pages/Profile/Profile.tsx
```diff
@@ -23,6 +25,24 @@ const SignOutButton = () => {
   )
 }
 
+const determineUserSkiSnowboardStatus = (
+  currentUserData?: UserAdditionalInfo
```
bcho892, line 29: also nit: prefer pass by value when possible (just including the skiier and snowboarding) so the function can't "accidentally" mutate the object

ganznz: Updated, let me know if that's better before I merge

bcho892 (approved):
lgtm

PR #408, opened 2024-06-08
388 delete user endpoint

ganznz:


server/src/middleware/routes.test.ts
```diff
@@ -334,6 +334,57 @@ describe("Endpoints", () => {
...
+        .set("Authorization", `Bearer ${adminToken}`)
+        .send({ uid: MEMBER_USER_UID })
+
+      expect(res.status).toEqual(200)
+      const deletedUser = await new UserDataService().getUserData(
+        MEMBER_USER_UID
+      )
+      expect(deletedUser).toEqual(undefined)
+      const unaffectedUser = await new UserDataService().getUserData(
+        GUEST_USER_UID
```
bcho892, line 361: we should check if the user is actually deleted in auth, as that will cause problems if they do

note the uid is same for both auth and firestore

bcho892: idk if the `AuthService` already has a method for that but if not can just use the `firebase-admin` package in the tests

server/src/service-layer/controllers/UserController.ts
```diff
@@ -56,4 +60,31 @@ export class UsersController extends Controller {
...
+        const authService = new AuthService()
+        const userDataService = new UserDataService()
+
+        const userClaims = await authService.getCustomerUserClaim(userUid)
+        if (userClaims.admin) {
+          this.setStatus(403) // forbidden request
+          return { error: "Cannot delete another admin." }
+        }
+
+        await authService.deleteUser(userUid)
```
bcho892, line 82: missing success code

ganznz:
im sorry benson idk what ive done and now codegen check is failing 🥺🥺🥺

server/src/middleware/__generated__/swagger.json
```diff
@@ -553,7 +577,7 @@
 					},
 					"nextCursor": {
 						"type": "string",
-						"description": "Needed for firestore operations which do not support offset\nbased pagination\n\n**Will be undefined in case of last page**"
+						"description": "Needed for firestore operations which do not support offset\r\nbased pagination\r\n\r\n**Will be undefined in case of last page**"
```
bcho892, line 580: Bruh it's doing this again, I'll try investigate why it does this but for now just need to remove all the `\r` (it's a Linux vs windows moment)

bcho892 (approved):
LGTM!

PR #415, opened 2024-06-10
372 fetch users booking between date

AzizPatel786:


server/src/data-layer/services/UserDataService.test.ts
```diff
@@ -1,6 +1,7 @@
 import { userInfoMock, userInfoMock2 } from "test-config/mocks/User.mock"
 import UserDataService from "./UserDataService"
 import { cleanFirestore } from "test-config/TestUtils"
+import * as console from "console"
```
bcho892, line 4: remove

server/src/data-layer/services/UserDataService.ts
```diff
@@ -1,5 +1,6 @@
 import FirestoreCollections from "data-layer/adapters/FirestoreCollections"
 import { UserAdditionalInfo } from "data-layer/models/firebase"
+// import * as console from "console"
```
bcho892, line 3: remove

server/src/middleware/routes.test.ts
```diff
@@ -704,4 +704,126 @@ describe("Endpoints", () => {
...
+          endDate
+        })
+
+      // console.log(res.body.users)
+      // console.log(res.body.date)
+      console.log(res.body.data[1])
+
+      expect(res.status).toEqual(200)
+      expect(res.body.data).toHaveLength(2)
+      // expect(res.body.data).toEqual(
```
bcho892, line 764: removing comments is recommended to avoid polluting codebase

server/src/service-layer/controllers/BookingController.ts
```diff
@@ -85,4 +92,62 @@ export class BookingController extends Controller {
...
+
+      console.log(responseData)
+
+      this.setStatus(200)
+
+      /** Returning the response data */
+      return { data: responseData }
+    } catch (e) {
+      this.setStatus(500)
+
```
bcho892, line 149: could `console.error` the error here

server/src/service-layer/response-models/CommonResponse.ts
```diff
@@ -1,4 +1,15 @@
+import { UserResponse } from "./UserResponse"
+import { Timestamp } from "firebase-admin/firestore"
+
 export interface CommonResponse {
   error?: string
   message?: string
 }
+
```
bcho892, line 8: this is in the wrong file, is not a common response

server/src/middleware/routes.test.ts
```diff
@@ -704,4 +704,126 @@ describe("Endpoints", () => {
...
+  describe("fetchUsersByBookingDateRange", () => {
+    beforeEach(async () => {
+      await createUsers()
+    })
+
+    afterEach(async () => {
+      await cleanFirestore()
+    })
+
+    afterAll(async () => {
```
bcho892, line 717: nit: not needed

server/src/middleware/routes.test.ts
```diff
@@ -704,4 +704,126 @@ describe("Endpoints", () => {
       expect(dates[0].date).toEqual(startDate)
     })
   })
+
+  describe("fetchUsersByBookingDateRange", () => {
```
bcho892, line 708: nit: more helpful to have `describe("<endpoint path>")`

bcho892 (changes requested):
test and logic looks good, just some minor issues. Also when you fix the merge conflicts i recommend:

- Accepting ALL changes (will make it easier to not delete other code during the merge)
- regenerating with `yarn dev-server` after

bcho892:
You might find #371 interesting, as it calls this endpoint on the frontend

server/src/service-layer/response-models/UserResponse.ts
```diff
@@ -14,11 +14,11 @@ type CombinedUserData = UserAdditionalInfo & {
   /**
    * The email the user uses to log in
    */
-  email: string
+  email?: string
```
bcho892, line 17: id recommend not making this change in here as this this might be enforced for a reason

bcho892: they are undefined (and giving errors) because the email and membership are handled separately than in the `UserDataService`, what you would need to do is use the function `bulkRetrieveUsersByUid` in `AuthService` to get that data,

see the `getAllUsers` method in `AdminController` for how the email is retrieved.

bcho892: also small tip, you can do `Partial<{key1:string}>` if you want to make **ALL** fields of a type optional

bcho892 (approved):
Otherwise lgtm, it doesn't seem to break the frontend so im ok with it for now 🤷

server/src/data-layer/services/UserDataService.ts
```diff
@@ -1,5 +1,11 @@
 import FirestoreCollections from "data-layer/adapters/FirestoreCollections"
 import { UserAdditionalInfo } from "data-layer/models/firebase"
+import AuthService from "../../business-layer/services/AuthService"
```
bcho892, line 3: ah you should do this inside your endpoint not the service

bcho892: its isolated into `AuthService` so it can be independently tested

PR #416, opened 2024-06-11
HOT: add script to deal with tsoa swagger generation

bcho892:
The workflow `codegen-check` runs on an ubuntu machine, which means the diff generated will cause it to fail on **correct** windows generated docs if there are newlines in strings. 

This PR introduces a new script that is called with `yarn dev-server` (the recommended way to regenerate types) which automatically removes the `\r` characters.


asun555 (approved):
LGTM

Issue #284, opened 2024-04-24
[FRONTEND] Add store action to refresh the current users id token


bcho892:
**Is your feature request related to a problem? Please describe.**
We may need a way to force refresh the id token (likely after the user has paid) so we can get the up-to-date claims without the user needing to refresh the page. This can be done using the [`.getIdToken(true)`](https://firebase.google.com/docs/reference/js/v8/firebase.User#getidtoken) function

You will need to edit `client\src\store\store.ts` and handle cases where the user is logged out etc. Refer to the [Sweet State docs on actions](https://atlassian.github.io/react-sweet-state/#/basics/actions)

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #417, opened 2024-06-11
Refresh user tokens on pages that require it

bcho892:
Have called the store action created in #284 on the:
- `bookings` 
- `profile`
pages so that the user can see up to date information about the views.

The user may now *not* navigate away from the payment confirmation page until their account has been logged back in by firebase


asun555 (approved):
LGTM

Issue #419, opened 2024-06-11
[FRONTEND] Make admin availability display the available dates


bcho892:
**Is your feature request related to a problem? Please describe.**

Right now on the availability page it displays "open" for available dates (shown below)

You will have to edit `client\src\components\composite\Admin\AdminAvailabilityView\AdminAvailabilityView.tsx`

![Image](https://github.com/UoaWDCC/uasc-web/assets/100653148/1b86ccfc-a5c0-430e-b9c1-bfd71ac05d9c)

Instead, we want it to actually show the availabilities:

![Image](https://github.com/UoaWDCC/uasc-web/assets/100653148/9998d14f-82d7-4470-8d05-cc0314c0e34b)

It will just involve doing something similar to #288 

You should also replace the manual timestamp conversions with the function `timestampToDate` in `client\src\components\utils\Utils.tsx`

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #421, opened 2024-06-11
HOT: check for dates not being included

bcho892:
There was a bug before where the user could bypass our validation by using the inputs to select dates that weren't available on the calendar. This new change just makes sure that the selected dates need to have a corresponding booking slot

Also created a helper function for converting dates that should be consumed in #419


eddiew16597 (approved):
looks good 👍🏻

PR #423, opened 2024-06-11
246 excitement slider

rxia357:


bcho892 (approved):
I'd recommend only using tailwind for styles that don't change because it's faster but other than that lgtm

bcho892:
Also link the issue that this closes in the "development" section on the right

rxia357:
thanks for the comments benson. I managed to use tailwind for most parts, however the clippath and the thumb styling are more complicated, so I just kept it as it is. I will merge this for now, and if you want any further adjustments I'm happy to change them.

PR #429, opened 2024-06-13
Add confirmation before user selects membership type

bcho892:
Now the membership type is not changed if they don't agree with the description of it


asun555 (approved):
LGTM

PR #430, opened 2024-06-13
fix overflow issues created by pricing cards

bcho892:
The home page was being overflowed by the pricing cards - with the payment screen also having some weird overflow issues. This ticket fixes those


asun555 (approved):
LGTM

PR #433, opened 2024-06-13
Booking payment endpoint frontend improvements

bcho892:
The frontend needed some more details from the backend such as:

- Retrieving existing sessions automatically
- Knowing type of error

This PR makes some small improvements so payment can be handled more smoothly on the front end

Issue #337, opened 2024-05-06
[FRONTEND] Implement edit functionality on profile page


bcho892:
**Is your feature request related to a problem? Please describe.**

Blocked by #336 

You will need to pass in edit handlers to the component created in #327 

[A mutation hook](https://tanstack.com/query/v4/docs/framework/react/reference/useMutation) also needs to be created that calls the `edit-self` endpoint

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

bcho892:
Already done in #509 and #520

PR #436, opened 2024-06-14
Allow for booking creation on frontend

bcho892:
# Working booking creation on front end 🤡 

## Page layout

image

## Basic booking creation flow
bookingpayment

## If there is an existing session

existingsession

## Conflict (the user has already booked)

bookingconflict

Makes use of the changes in #433 for backend that allows existing sessions to be fetched, however there have been no logic changes.

Have also created some hooks that allow for certain actions to be performed on application state:

- `useIsUserLoggedIn`, alternative to using `useAppData`
- `useUserLoggedInCallback`, allows effects to be run when the auth state changes from logged out to logged in

Also implemented a mutation hook for `editSelf` which can be consumed in #337


asun555 (approved):
LGTM

PR #441, opened 2024-06-14
Send additional password reset email on sign up

bcho892:
image

Before the only way to change password was after the payment - now an additional method has been added in case the user doesn't reach the account setup stage.


asun555 (approved):
LGTM

PR #446, opened 2024-06-15
Improve date handling on frontend

bcho892:
The problem before is the different timezones on the frontend return different dates from the date inputs. The fix here is to use `getDay` on the front end (to display the appropriate data to the user) and backend **but** on the back end need to re-parse the date to its timezone to account for differences. 

- Switch to getDay instead of getUTCDay (helps to resolve the issue of dates getting "pushed back" one day)
- Add check on frontend to disallow selection of 10 > nights
- Display the dates the current booking is for in the checkout session


asun555 (approved):
LGTM

Issue #426, opened 2024-06-12
[FRONTEND] Create add booking popup


bcho892:
**Is your feature request related to a problem? Please describe.**

![Image](https://github.com/UoaWDCC/uasc-web/assets/100653148/52428cf6-42e8-414f-86d7-2d4bb3c2ff00)

Will need to call the endpoint to manually create a booking

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #458, opened 2024-06-16
allow admins to create new user on frontend

bcho892:
The use case is that sometimes admin may require manual user signup in case of non-stripe payments such as bank etc. This means that they should have the option to add members via the `AdminMemberView`. 

image

image

Also create a generic modal container called `ModalContainer` which can possibly be used in #426


asun555 (approved):
LGTM

PR #462, opened 2024-06-17
Allow admins to delete users

bcho892:
Had to make some changes to the endpoint made in #408, because in some cases we still want to carry out all the steps even if the previous ones failed. Also introduced an `isUpdating` prop to give feedback that a deletion is happening

delete users


asun555 (approved):
LGTM

PR #463, opened 2024-06-17
HOT: remove bad normalise

bcho892:
Since we already do it on the front end, the extra processing on the back end makes the date "later"


eddiew16597 (approved):
approved

PR #464, opened 2024-06-18
replace for loop with Promise.all in fetch-users endpoint

bcho892:
This does not change any logic, just means that we can take advantage of concurrency and improve performance


asun555 (approved):
LGTM

Issue #431, opened 2024-06-13
[FRONTEND] Investigate if there is a way to avoid "normalising" dates before sending to backend


bcho892:
**Is your feature request related to a problem? Please describe.**

Right now in #389 we made the decision to convert dates to a date string and back into a timestamp to ensure consistent time information. However it would be preferred if there is a better way. 

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #465, opened 2024-06-18
format dates properly for display

bcho892:
The changes in #463 meant that while the bookings were functionally correct, the displayed dates to the user were wrong. This is why we have to normalise the dates to make the correct ones displayed (to be consistent with the ones in admin view - #431 for further discussion)


asun555 (approved):
LGTM

eddiew16597 (commented):
looks good

PR #468, opened 2024-06-20
add optional ethnicity field to sign up form & reset sign up form on successful sign up

bcho892:
image

Short change that just involves updating the models and adding a text input.

An extra feature added is to reset the sign up form once the sign up is successful


asun555 (approved):
LGTM

PR #483, opened 2024-06-21
Redirect user on password reset success

bcho892:
Before it would just stay on the password reset form, now itll take them to the success page


asun555 (approved):
LGTM

PR #485, opened 2024-06-21
Fix date of birth input crashing site

bcho892:
Before we weren't checking if the value was undefined, also added appropriate limits to ages


asun555 (approved):
LGTM

PR #489, opened 2024-06-21
Fix typo in prod server workflow, and update the environments associated with each

bcho892:
The prod deployment wouldn't run due to missing `}` and also remade the environements to be more specific


asun555 (approved):
LGTM

Issue #475, opened 2024-06-20
[BACKEND] Add functionality to booking creation popup and integrate into page


bcho892:
**Is your feature request related to a problem? Please describe.**

Following the work done in #473 we should make it call the actual endpoint (in #380) and integrate it on the front end.

**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] [Schemas annotated](https://tsoa-community.github.io/docs/annotations.html) if adding new models
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

Issue #459, opened 2024-06-17
[BACKEND] Use Promise.all instead of for loop in fetch-users


bcho892:
**Is your feature request related to a problem? Please describe.**

introduced in #415 was a method to fetch users between a date. while doing #457 we notice there was super slow querying on larger date ranges, and it became apparent that it was because we were doing blocking async calls in a for loop, not taking advantage of concurrency. 

We should replace it with a `Promise.all`

Further reading: https://stackoverflow.com/questions/53798589/await-loop-vs-promise-all

**BEFORE MERGING**
- [ ] Integration test written for services
- [ ] [Schemas annotated](https://tsoa-community.github.io/docs/annotations.html) if adding new models
- [ ] Code generation run (*hint*: `yarn workspace server tsoa spec-and-routes`)
- [ ] Appropriate mocks created where possible
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (*hint*: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #508, opened 2024-06-24
380 backend create endpoint to create bookings

Kartik-M24:


server/src/service-layer/response-models/BookingResponse.ts
```diff
@@ -17,3 +17,11 @@ export interface UsersByDateRangeResponse {
   data?: Array<{ date: Timestamp; users: CombinedUserData[] }>
   error?: string
 }
+
+/**
+ * Represents the response structure for fetching user ids by date range.
+ */
+export interface UIdssByDateRangeResponse {
```
Kartik-M24, line 24: Just wanted to clarify if this was okay to do or was I meant to use another response?

server/src/middleware/routes.test.ts
```diff
@@ -1275,4 +1255,118 @@ describe("Endpoints", () => {
...
+  describe("/bookings/create-bookings", () => {
+    beforeEach(async () => {
+      await createUsers()
+    })
+
+    afterEach(async () => {
+      await cleanFirestore()
+    })
+
+    it("should return userIds for successful bookings within the date range", async () => {
```
bcho892, line 1287: we should add a check to directly query firestore and see if the bookings are actually being created properly

server/src/service-layer/controllers/BookingController.ts
```diff
@@ -32,6 +35,94 @@ import {
...
+            delete userIds[i] // Remove user from list if they already have a booking
+            break
+          } else {
+            bookingDataService.createBooking({
+              user_id: userIds[i],
+              booking_slot_id: slotIds[i],
+              stripe_payment_id: stripePaymentIds[i]
+            })
+          }
+        }
```
bcho892, line 96: refer to #459, this will get very slow for large data sets, nede to use `Promise.all`

server/src/service-layer/controllers/BookingController.ts
```diff
@@ -32,6 +35,94 @@ import {
...
+        let i: number
+        for (i = 0; i < userIds.length; i++) {
+          if (
+            (await bookingDataService.getBookingsByUserId(userIds[i]))
+              .length !== 0
+          ) {
+            delete userIds[i] // Remove user from list if they already have a booking
+            break
+          } else {
+            bookingDataService.createBooking({
```
bcho892, line 90: this needs to be async

server/src/service-layer/response-models/BookingResponse.ts
bcho892: yea should be fine

bcho892 (changes requested):
I've reverted all of the unneeded changes, but there are just some small changes that should be made for reliability

server/src/middleware/routes.test.ts
```diff
@@ -1275,4 +1274,139 @@ describe("Endpoints", () => {
       expect(res.status).toEqual(401)
     })
   })
+
+  describe("/bookings/create-bookings", () => {
```
bcho892, line 1278: nit: Like discussed before, we want to test cases we don't want to happen, such as a user being added to the booking twice - it would be nice to have a test case for this

bcho892 (approved):
LGTM, just a small improvement if you feel like doing it

bcho892:
also i think we forgot about something important in that ticket, in the request model needed to put the `userId`. I will fix this in #475

PR #389, opened 2024-06-03
booking payment endpoint

jeffplays2005:
# Endpoint to create checkout sessions for bookings

Bulk of work done by @jeffplays2005

This endpoint meets the requirements of:
- Validating that the dates are indeed available with regard for if users are already in sessions
- Returning existing sessions
- Serializing data in the correct format for the endpoint in #370 
- Max 10 dates bookable
- Start date min today, max 1 year in future

### Normal Date Range

bookingsendpoint

### Single Friday/Saturday
*Note the stripe product is incorrectly called weekend but the logic concerning it is the same*

singlefriday

PR #511, opened 2024-06-26
Minus off currently booked slots from availabilities

bcho892:
Right now it doesn't take this check into account, this mostly just copies the logic from #389. 

Also took initiative to display how long the user has to book, and also remind them of the dates they are booking


asun555 (approved):
LGTM

PR #513, opened 2024-06-26
backend webhook send email to user notifying them of their booking

bcho892:
Made use of https://medium.com/@y.mehnati_49486/how-to-send-an-email-from-your-gmail-account-with-nodemailer-837bf09a7628 to send directly from a gmail account using an app password. 

also updated workflows, will update wiki soon with the new envs

Also closes #461 

`NODE_MAILER_EMAIL`
and `NODE_MAILER_PASSWORD`

PR #518, opened 2024-06-26
Refactor to store session dates in session metadata

bcho892:
Made sure that we display the correct dates, moved all info to metadata of stripe, #513 needs to be redone now


asun555 (approved):
LGTM

PR #523, opened 2024-06-26
498 backend include booking id in fetch users endpoint

Kartik-M24:


server/src/service-layer/controllers/BookingController.ts
```diff
@@ -328,7 +334,11 @@ export class BookingController extends Controller {
         /** Adding the date and users to the response data array */
         responseData.push({
           date: slot.date,
-          users: combinedUsers
+          // Mapping the users to include the booking ID
+          users: combinedUsers.map((user, i) => ({
+            ...user,
+            bookingId: bookingSlotIds[i]
```
bcho892, line 340: We want to return the actual Id of the booking, instead of the booking slot id. Also a side note it's very brittle to be using indexes for something like this because you cant assume they are sorted. Would recommend the `find` method instead

bcho892 (changes requested):

Kartik-M24:
How do I delete the previous commits which are useless?

bcho892:
> How do I delete the previous commits which are useless?

There's not really any need if the diff is fine, but what you can do is reset to master then cherry pick commits

bcho892 (approved):

server/src/service-layer/response-models/UserResponse.ts
```diff
@@ -26,3 +26,7 @@ export interface AllUsersResponse
     CursorPaginatedResponse {
   data?: CombinedUserData[]
 }
+
+export interface BookingIdandUserData extends CombinedUserData {
```
bcho892, line 30: nit: can possibly annotate this one (so when we generate code others have documentation) but dont worry about it for now

PR #537, opened 2024-06-28
Allow existing sessions to be successfully found

bcho892:
Before it was trying to parse undefined dates, now move that to after confirming there's an existing one


zlrkw11 (approved):
looking real good(I have no idea what this ticket means)

PR #540, opened 2024-06-28
fix cock blocking on admin booking view

bcho892:
just made it so that the modal contents are actually hidden when we close it


eddieyw (approved):
Nice 1 line change

PR #545, opened 2024-06-28
add error boundary for components

bcho892:
Now when an error is thrown the user will be told to refresh the page instead of getting a blank screen.


asun555 (approved):
LGTM

PR #551, opened 2024-06-28
442 block payment endpoint if email !verified

AzizPatel786:


bcho892:
Looks good, just need to make some other changes (on frontend etc) before this is safe to merge in

bcho892:
I think this PR is not needed anymore, as most of the time this problem will be manually resolved anyway. Thanks for working on it though!

PR #552, opened 2024-06-29
allow admins to delete bookings on front end

bcho892:
Called the endpoint made in #408 

Also think we need to make the think a delete endpoint

https://github.com/UoaWDCC/uasc-web/assets/100653148/d5ee8ec8-33ed-4e99-8a13-fd15e283c7ed


eddieyw (approved):
approved.

PR #556, opened 2024-06-29
530 expire checkout sessions last spot

AzizPatel786:


bcho892:
Ok, I think I should give some background on how all the dates work on stripe:

We have `booking_slots` which refer to the date and has a `max_bookings`. The way we calculate the availability is:

1. check how many slots there are in the `bookings` collection - which refers to individual bookings made, **no** date data is stored in there, instead it has a field that refers to an `id` of a `booking_slot`
2. the available slots for a `booking_slot` is: the `max_bookings` **minus** the amount of `bookings` pointing to a `booking_slot` 
3. in case of displaying availability to the user we also minus off the amount of checkout sessions that have that `booking_slot`'s id in its metadata

The way we use this with stripe is when we create a checkout **session** we add an array of dates pointing to the `booking_slot` id (serialised as a string with `JSON.stringify`) to the metadata.

You will have to get the checkout session and `JSON.parse` the metadata to find the relevant checkout session ids that include the now fully booked dates.

server/src/service-layer/controllers/StripeWebhook.ts
```diff
@@ -69,6 +71,14 @@ export class StripeWebhook extends Controller {
...
               try {
                 await stripeService.handleBookingPaymentSession(uid, session)
+
+                // Check if the last spot is taken and expire other sessions
+                const date = session.metadata.date
+                const isLastSpot =
+                  await bookingSlotService.isLastSpotTaken(date)
+                if (isLastSpot) {
+                  await this.expireOtherCheckoutSessions(stripe, date)
+                }
```
bcho892, line 81: this should be in `handleBookingPaymentSession`

bcho892: this means you already have access to the session

server/src/data-layer/services/BookingSlotsService.ts
```diff
@@ -73,4 +73,12 @@ export default class BookingSlotService {
...
     return await FirestoreCollections.bookingSlots.doc(bookingSlotId).delete()
   }
+
+  public async isLastSpotTaken(date: string): Promise<boolean> {
+    const bookingSlots = await FirestoreCollections.bookingSlots
+      .where("date", "==", date)
+      .get()
+    const availableSlots = bookingSlots.docs.length
+    return availableSlots <= 1 // Current booking is the last one
+  }
```
bcho892, line 83: probably not needed, also like stated in the previous comment, there will only every be **one** booking slot for a date

bcho892 (changes requested):

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
...
+      expand: ["data.payment_intent"]
+    })
+
+    // Pending checkout sessions for this slot
+    const pendingSessions = sessions.data.filter(
+      (session) =>
+        session.metadata.booking_slot_id === bookingSlotId &&
+        (session.payment_intent as Stripe.PaymentIntent)?.status ===
+          "requires_payment_method"
+    ).length
```
bcho892, line 489: don't need this, will always be 0 because NZ doesn't accept async payment

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
...
+        limit: 100,
+        expand: ["data.payment_intent"]
+      })
+
+      const sessionsToExpire = sessions.data.filter(
+        (session) =>
+          session.metadata.dates.includes(date) &&
+          (session.payment_intent as Stripe.PaymentIntent)?.status ===
+            "requires_payment_method"
+      )
```
bcho892, line 512: You should look at some similar (but still quite different) logic (from `server\src\service-layer\controllers\PaymentController.ts`) 

```ts
      const openSessions = await stripeService.getRecentActiveSessions(
        CheckoutTypeValues.BOOKING,
        MINUTES_AGO,
        true
      )

// example if i want to check what booking slots it involves
const firstSessionBookingSlots = JSON.parse(openSessions[0].metadata[BOOKING_SLOTS_KEY]) as Array<string>

// if that contains the now fully booked booking slot then we expire the session.
```

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
...
+    try {
+      // Fetch all checkout sessions for the specific date
+      const sessions = await stripe.checkout.sessions.list({
+        limit: 100,
+        expand: ["data.payment_intent"]
+      })
+
+      const sessionsToExpire = sessions.data.filter(
+        (session) =>
+          session.metadata.dates.includes(date) &&
```
bcho892, line 509: metadata can *only* be stored as a string, hence you will need to use `JSON.parse`

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
...
+          "requires_payment_method"
+    ).length
+
+    const availableSlots =
+      bookingSlot.max_bookings - bookingCount - pendingSessions
+    return availableSlots <= 0
+  }
+
+  private async expireOtherCheckoutSessions(
+    stripe: Stripe,
```
bcho892, line 497: is this not global?

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
...
+    const bookingSlot =
+      await bookingSlotService.getBookingSlotById(bookingSlotId)
+
+    const bookings = await bookingDataService.getBookingsBySlotId(bookingSlotId)
+    const bookingCount = bookings.length
+
+    const sessions = await stripe.checkout.sessions.list({
+      limit: 100,
+      expand: ["data.payment_intent"]
+    })
```
bcho892, line 481: not needed

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +463,71 @@ export default class StripeService {
     }
   }
 
+  private async isLastSpotTaken(
+    bookingSlotService: BookingSlotService,
+    bookingDataService: BookingDataService,
+    stripe: Stripe,
```
bcho892, line 469: we can just initialise this in the function, don't need to pass in

bcho892 (commented):
getting closer

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +458,58 @@ export default class StripeService {
...
+        expand: ["data.payment_intent"]
+      })
+
+      const sessionsToExpire = sessions.data.filter((session) => {
+        const bookingSlots = JSON.parse(
+          session.metadata[BOOKING_SLOTS_KEY]
+        ) as Array<string>
+        return (
+          bookingSlots.includes(bookingSlotId) &&
+          (session.payment_intent as Stripe.PaymentIntent)?.status ===
```
bcho892, line 491: Not needed

bcho892: This cast is also pretty dangerous

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +458,58 @@ export default class StripeService {
...
+    const availableSlots = bookingSlot.max_bookings - bookingCount
+    return availableSlots <= 0
+  }
+
+  private async expireOtherCheckoutSessions(
+    bookingSlotId: string
+  ): Promise<void> {
+    try {
+      // Fetch all checkout sessions for the specific booking slot id
+      const sessions = await stripe.checkout.sessions.list({
```
bcho892, line 480: Suggest using the relevant method in stripeservice and get all sessions from 24 hrs (max) ago

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +458,58 @@ export default class StripeService {
...
+        return (
+          bookingSlots.includes(bookingSlotId) &&
+          (session.payment_intent as Stripe.PaymentIntent)?.status ===
+            "requires_payment_method"
+        )
+      })
+
+      // Expire each session that matches the booking slot id
+      await Promise.all(
+        sessionsToExpire.map(async (session) => {
```
bcho892, line 498: Nice

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +458,58 @@ export default class StripeService {
...
+        limit: 100,
+        expand: ["data.payment_intent"]
+      })
+
+      const sessionsToExpire = sessions.data.filter((session) => {
+        const bookingSlots = JSON.parse(
+          session.metadata[BOOKING_SLOTS_KEY]
+        ) as Array<string>
+        return (
+          bookingSlots.includes(bookingSlotId) &&
```
bcho892, line 490: Nice

bcho892:
Good job,pretty much the right logic,

bcho892 (approved):
nice!

server/src/business-layer/services/StripeService.ts
```diff
@@ -452,6 +458,54 @@ export default class StripeService {
     }
   }
 
+  private async isLastSpotTaken(bookingSlotId: string): Promise<boolean> {
```
bcho892, line 461: also we can definitely write a test for this, if we move it to bookingslotservice

bcho892: I'd HIGHLY recommend doing this btw since we are going based off a code review only

bcho892 (approved):
Good job on seeing this through

PR #559, opened 2024-06-29
Improve footer links

bcho892:
Before there were a lot of dead links. Now also keep all of them for mobile as well.


asun555 (approved):
LGTM

PR #560, opened 2024-06-29
set up analytics

bcho892:
added a few points where analytics may be interesting, but mostly just getting the framework set up.

Also defined `vite-env.d.ts` to allow for intellisense


asun555 (approved):
LGTM

PR #569, opened 2024-07-02
Make all workflows use volta and then remove corepack usage

bcho892:
Volta allows us to just change the `package.json` which makes it easier to update our dev dependencies. Making this change means we don't have to deal with nvm etc


Dockerfile
```diff
@@ -1,23 +1,37 @@
 # syntax = docker/dockerfile:1
-# Adjust NODE_VERSION as desired
-ARG NODE_VERSION=20
-FROM node:${NODE_VERSION}-slim as base
+FROM debian:stable-slim as base
```
bcho892, line 2: using https://dev.to/michalbryxi/volta-in-docker-162a

.github/workflows/lint.yml
```diff
@@ -10,14 +10,14 @@ jobs:
...
-      - name: Enable corepack
-        run: corepack enable
+      - name: Install volta
+        uses: volta-cli/action@v4
 
       - name: Checkout repository
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
 
       - name: Install NPM dependencies
-        run: sudo yarn
```
bcho892, line 20: `sudo` was forcing it to use the default `yarn`

asun555 (approved):
LGTM

PR #575, opened 2024-07-02
Allow sorting users by role on admin member view

bcho892:
For the UI design just decided to use a button that cycles through states

sorting


asun555 (approved):
LGTM

PR #584, opened 2024-07-04
Added policy to booking page but encountered error

DRFRezze:


bcho892 (approved):
LGTM! just fix the import error

PR #585, opened 2024-07-04
put booking policy on bookings page

DRFRezze:


bcho892 (approved):
ok it looks like it was git being dumb have fixed it

PR #589, opened 2024-07-05
Confirm the users dates (actual check in and check out) on bookings page

bcho892:
The client is worried about people misinterpreting the dates they are booking, so we have added an extra confirmation

image


Kartik-M24 (approved):
LGTM

PR #592, opened 2024-07-06
Migrate project to nextjs

bcho892:
# Migration to NEXT.JS

This is a big change, but hopefully it makes this project more future-proof and serves the client better.

## What's going to happen to `/client`?

We will still keep the folder temporarily, however we will rename `uasc-next` to `client` and completely get rid of the old vite app once the new version is stable. #593 

## Next.js config
We are using [**Next.js 14.2**](https://nextjs.org/blog/next-14-2) with [app router](https://nextjs.org/docs/app). App router seems to work pretty well with what we have (good amount of static content and SPA)

What changing to next.js does is:
- Offer easy [SSR](https://prismic.io/blog/client-side-vs-server-side-rendering) and meta tag config if we need it
- Allow us to have better code-splitting (as opposed to the hand rolled `lazy`)
- [Turbopack](https://turbo.build/pack)
- A structured way of building the routes with app router

We likely do not need to use API routes, but there is the option should we require it

## Deployment changes

Due to the free-tier limits on netlify we need to be conservative when building. For this reason all builds on netlify should **only** be for changes to production. For credentials to the account please contact relevant WDCC execs. This will likely be given the uasc.co.nz domain

> Deployments to staging will remain on https://uasc-ceebc.web.app/ as a statically built next.js site. 

## Differences with the vite app

Apart from the page structure, one can pretty much keep writing "normal" react. The only exception is if you want to [dynamically fetch data](https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration#static-site-generation-getstaticprops). 

One thing to note is that there will be **no** `react-router-dom` anymore, and you will have to look at the alternatives from `next/navigation` (not to be confused with `next/router` which should **not** be used.


eddieyw (approved):
Code looks good for nextjs migration

PR #595, opened 2024-07-07
change build config to use trailing slash and setup prod deployments on GH pages

bcho892:
Before without using trailing slashes the static site failed to update its tabs properly and couldnt handle dynamic routes (see https://nextjs.org/docs/pages/building-your-application/deploying/static-exports#deploying). 

Also set up the prod deployment for gh pages, seems to have better bandwidth (100GB bandwidth vs 10GB for firebase) and a more generous build limit compared to netlify (unlimited and allows 10 per hour compared to 300 minutes per month)

also fixed a double image bug


asun555 (approved):
LGTM

PR #597, opened 2024-07-07
Fix gh pages deployment issues

bcho892:
- Solved problem of css `url` not properly being appended with the base path
- Changed `a` tags to use the native next `Link`


asun555 (approved):
LGTM

PR #604, opened 2024-07-08
display placeholder on protected routes to improve UX

bcho892:
Now instead of flashing the admin/profile page, itll show the user a redirect before


asun555 (approved):
LGTM

PR #617, opened 2024-07-12
593 frontend replace client directory with new nextjs frontend

bcho892:
Deleted old client directory and rename all workflows. This means that we deploy everything using the static site generation by next


zlrkw11 (approved):
Looking clean

PR #626, opened 2024-07-13
Setup sanity as a cms in project

bcho892:
We got it to this screen:

image

Afterwards have to add schemas in future tickets

The setup was straightforward, mostly following https://www.sanity.io/plugins/next-sanity 

New sanity project has been created (with two datasets: `staging` and `production`), and workflows updated accordingly

Changed some stories because ts now recognising to throw too complex type union error for the button props unions


zlrkw11 (approved):
Looking clean

PR #630, opened 2024-07-13
Connect About sections to sanity.io

bcho892:
image

Had to bump sanity version based on error in https://github.com/sanity-io/next-sanity/issues/1306#issuecomment-2072899677 

Also defined format that we should follow for new schemas in future:

Under `client/src/models/sanity/[SchemaName]/Schema` need to export:
- Schema for sanity client (using `defineField`)

Under `client/src/models/sanity/[SchemaName]/Utils` need to export:
- ts type (for static fetching)
- Relevant queries for that schema (for our use cases its pretty much only going to be getting all of them)
   - cheat sheet: https://www.sanity.io/docs/query-cheat-sheet

> This extra file exists to avoid dependency on sanity in the about page which was causing build errors


.github/workflows/firebase-client-hosting-merge.yml
```diff
@@ -16,7 +16,7 @@ env:
   NEXT_PUBLIC_BACKEND_BASE_URL: ${{secrets.VITE_BACKEND_BASE_URL}}
   NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY: ${{secrets.TEST_STRIPE_PK}}
   NEXT_PUBLIC_SANITY_PROJECT_ID: ${{secrets.STAGING_SANITY_PROJECT_ID}}
-  NEXT_PUBLIC_SANITY_DATASET: staging
+  NEXT_PUBLIC_SANITY_DATASET: production
```
bcho892, line 19: This is ok because we want to be able to test changes before rolling out to prod

asun555 (approved):
LGTM

PR #632, opened 2024-07-14
make home page connected to CMS

bcho892:
image

Could add customisable youtube video, but probably not needed rn

Will need to restyle the member benefit cards to use images instead of the icons later


asun555 (approved):
LGTM

PR #641, opened 2024-07-17
639 Add Cron Workflow to Deploy Stable

jeffplays2005:
Added the new workflow and enabled cron schedule to update stable branch to the production app.
Removed the cron schedule from the master workflow, can only manually update production through master.


jeffplays2005:
Also created stable branch

bcho892 (approved):

PR #642, opened 2024-07-17
640 neaten endpoint tests

jeffplays2005:
# Update UserController route tests

Changes to: `/user/edit-self`
* Remove `should edit the user information for multiple attributes` test as can can combine with `should edit the users information`
* Removed `beforeEach` and `afterEach` from individual `describe` blocks and replaced with a general `beforeEach` and `afterEach` in the main `describe` block. This is per the [jest documentations](https://jestjs.io/docs/setup-teardown#scoping)


bcho892 (approved):

jeffplays2005:
Sorry, might just need another review!

server/src/middleware/tests/AdminController.test.ts
```diff
@@ -503,22 +430,14 @@ describe("AdminController endpoint tests", () => {
 
       expect(dates).toHaveLength(2)
       expect(dates[0].max_bookings).toBeLessThanOrEqual(0)
-      expect(dates[0].description).toEqual("my test")
       expect(removeUnderscoresFromTimestamp(dates[0].date)).toEqual(startDate)
 
       expect(dates[1].max_bookings).toBeLessThanOrEqual(0)
-      expect(dates[1].description).toEqual("skipped a date")
```
bcho892, line 510: why are these deleted?

jeffplays2005: Oh whoops, I saw an earlier test and thought that it would be working for every case afterwards 😬

bcho892 (approved):

PR #643, opened 2024-07-18
607 create an endpoint to demote non admin users

jeffplays2005:
Created endpoint in AdminController
Created new method in AuthService and its corresponding test, this method fetches all users(1000+)

Created endpoint tests, the endpoint isn't tested on 1000+ members as the fetch users method is confirmed to be working with 1000+ in the tests for AuthService.


jeffplays2005:
Screenshot 2024-07-18 at 4 47 53 PM

Will be removing the test for 1000+ users as this is somewhat redundant after testing locally and takes too long.

server/src/middleware/tests/AdminController.test.ts
```diff
@@ -156,6 +159,38 @@ describe("AdminController endpoint tests", () => {
...
+        .delete("/admin/users/demote-all")
+        .set("Authorization", `Bearer ${adminToken}`)
+        .send({})
+      expect(res.status).toEqual(200)
+
+      const { admin } = await authService.getCustomerUserClaim(ADMIN_USER_UID)
+      expect(admin).toEqual(true)
+    })
+    it("Should demote all members", async () => {
+      // Note that this isn't testing on creating over a thousand members
```
bcho892, line 176: yea should never do that

server/src/service-layer/controllers/AdminController.ts
```diff
@@ -361,6 +363,26 @@ export class AdminController extends Controller {
     }
   }
 
+  @SuccessResponse("200", "Demoted all non-admin users")
+  @Delete("/users/demote-all")
```
bcho892, line 367: Nit: this should be a patch

bcho892 (approved):

bcho892:
> Screenshot 2024-07-18 at 4 47 53 PM
> Will be removing the test for 1000+ users as this is somewhat redundant after testing locally and takes too long.

yea very bad idea to do this, what you could do is see if you can set a limit on the pagination and then make it configurable

server/src/service-layer/controllers/AdminController.ts
jeffplays2005: Oops, yep updated this

bcho892 (approved):

PR #647, opened 2024-07-18
611 Signup validation

jeffplays2005:
Require at least 3 items split by a comma.


client/src/store/SignUpForm.tsx
```diff
@@ -67,6 +68,13 @@ const actions = {
         if (phone_number.toString().length < 6) {
           invalidFields.push("Phone Number")
         }
+        const emergencyContactFields = emergency_contact?.split(",")
```
bcho892, line 71: This is too strict, I'd say just make sure it's not empty

bcho892 (approved):

PR #655, opened 2024-07-19
Add check in and check out dates to booking success page

bcho892:
# With valid dates
image

# Doesn't crash with invalid dates
image


asun555 (approved):
LGTM

bcho892:
Not going to work because 31, 30 will overflow -> not good.

bcho892:
Should handle this on backend

PR #665, opened 2024-07-21
add robots and sitemap generation

bcho892:
this is important because we are going to be using a domain likely to be indexed by google.

see:
- https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots
- https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap

We make sure that these only allow indexing if 

- `NEXT_CONFIG_ENV`
- `FRONTEND_BASE_URL`

are set

### DIFF
--- a/client/environment.d.ts
+++ b/client/environment.d.ts
@@ -21,12 +21,12 @@ declare global {
       readonly NEXT_PUBLIC_ENV_STRIPE_PUBLISHABLE_KEY: string
       /** Used for deployments  */
       readonly NEXT_CONFIG_ENV?: "production" | "staging"
-      /** What all paths should be prefixed with */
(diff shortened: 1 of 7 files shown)


asun555 (approved):
LGTM

client/environment.d.ts
```diff
@@ -21,12 +21,12 @@ declare global {
       readonly NEXT_PUBLIC_ENV_STRIPE_PUBLISHABLE_KEY: string
       /** Used for deployments  */
       readonly NEXT_CONFIG_ENV?: "production" | "staging"
-      /** What all paths should be prefixed with */
-      readonly DEPLOYMENT_BASE_PATH?: string
```
bcho892, line 25: not needed, was a hack solution for the old `gh-pages` deployment

PR #680, opened 2024-07-23
566 Delete preview channel upon merge

jeffplays2005:
Have added the feature to the pre-existing workflow action as per [this pull request on action-hosting-deploy](https://github.com/FirebaseExtended/action-hosting-deploy/pull/65)


jeffplays2005:
If the above doesn't work, I found another solution:
https://github.com/FirebaseExtended/action-hosting-deploy/issues/60#issuecomment-1761357523

bcho892:
> If the above doesn't work, I found another solution: [FirebaseExtended/action-hosting-deploy#60 (comment)](https://github.com/FirebaseExtended/action-hosting-deploy/issues/60#issuecomment-1761357523)

Prefer this as the current change seems to not have been implemented...

jeffplays2005:
> > If the above doesn't work, I found another solution: [FirebaseExtended/action-hosting-deploy#60 (comment)](https://github.com/FirebaseExtended/action-hosting-deploy/issues/60#issuecomment-1761357523)
> 
> Prefer this as the current change seems to not have been implemented...

Oops, what am I doing... Will make the changes and push soon

bcho892 (approved):
I don't think this will mess anything up because its our staging instance anyway

PR #682, opened 2024-07-25
566 delete preview channel on merge

jeffplays2005:
Have created another pull request to reflect the following issue:
Screenshot 2024-07-25 at 3 32 39 PM

The issue was caused by the deployment only using the first 20 characters of the pull request title and also the ticket number.


jeffplays2005:
Have made some changes to this ticket as have just realised that `FirebaseExtended/action-hosting-deploy@v0` [uses the format](https://github.com/FirebaseExtended/action-hosting-deploy/blob/7a831e3ee9349aecd334c2cc01080a2c71a29509/src/getChannelId.ts#L25):

```
pr<pull request ticket number>-<pull request commit title, substringed to the first 20 characters>
```
Screenshot 2024-07-25 at 3 30 43 PM

.github/workflows/delete-firebase-client-on-close.yml
```diff
@@ -13,7 +13,9 @@ jobs:
 
       - uses: w9jds/firebase-action@v12.4.0
         with:
-          args: hosting:channel:delete ${{ github.head_ref }} --force
+          # as we didn't specify a specific channel id, firebase-action defaults
+          # with using pr<ticket number>-<ticket reference substringed to 20 chars>
+          args: hosting:channel:delete pr${{ github.event.pull_request.number }}-${{ github.head_ref:0:20 }} --force
```
bcho892, line 18: have you tested that the string slicing works in bash?

jeffplays2005:
Found a workaround for the substring thing with yaml.
Github provides a output system and can use a shell workaround by using `cut` and saving the output.

https://github.com/orgs/community/discussions/26625#discussioncomment-5785675

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter

bcho892 (approved):
Ok i've tested it out and seems to work with the slicing

PR #689, opened 2024-07-26
Use pug for templating and add new template for bookings

bcho892:
# Result

image

Wiki at https://github.com/UoaWDCC/uasc-web/wiki/Sending-and-templating-emails has been updated.


asun555 (approved):
LGTM

PR #690, opened 2024-07-26
Document generic component props

bcho892:
avoided doing ones that were not in use (they probably need to be deleted in future tickets)


asun555 (approved):
LGTM

PR #691, opened 2024-07-26
copy templates to build outputs

bcho892:
Right now the backend fails to use the template because it can't compile

Also made the compile function last the app lifetime, do not need to be constantly recompiling.


asun555 (approved):
LGTM

PR #692, opened 2024-07-26
Create interface in firebase.ts

jeffplays2005:


jeffplays2005:
Essentially just an initial concept for the event booking slots, has room for improvement and expenditures.

server/src/data-layer/models/firebase.ts
```diff
@@ -129,3 +129,36 @@ export interface BookingChange {
...
+  /**
+   * The location of this event
+   */
+  location: string
+  /**
+   * The start date and end date of this event.
+   * Note that this date is UTC time.
+   * Use the same start and end date to show that its a 1 day event.
+   */
+  start_date: Timestamp
```
bcho892, line 157: each of these need to be individually annotated

server/src/data-layer/models/firebase.ts
```diff
@@ -129,3 +129,36 @@ export interface BookingChange {
...
+  /**
+   * The title of this event
+   * @example Snowboarding race!
+   */
+  title: string
+  /**
+   * An optional description for this event
+   * @example Come join your friends in this exciting UASC event and race!
+   */
+  description?: string
```
bcho892, line 143: maybe mention this should be markdown?

bcho892 (approved):
happy for merge if wiki is updated

jeffplays2005:
Wiki created 

https://github.com/UoaWDCC/uasc-web/wiki/Firestore-Architecture#event_slot-collection

PR #695, opened 2024-07-27
Add interface for EventReservation in firebase.ts

jeffplays2005:
Will add into the wiki after approval


server/src/data-layer/models/firebase.ts
```diff
@@ -167,3 +167,27 @@ export interface EventSlot {
...
+  email: string
+  /**
+   * Boolean to check if the user is a member
+   * @example true
+   */
+  is_member: boolean
+  /**
+   * A link from EventReservation -> EventSlot
+   */
+  event_slot: string
```
bcho892, line 192: I was thinking that we could make this a subcollection instead, that might simplify it? compared to what we did for bookings

server/src/data-layer/models/firebase.ts
```diff
@@ -166,4 +186,9 @@ export interface EventSlot {
    * @example 30
    */
   max_occupancy?: number
+  /**
+   * Subcollection of EventReservation interface.
+   * Holds the members that made a reservation to this event
+   */
+  reservations: EventReservation[]
```
bcho892, line 193: don't need this refer to what subcollections are in firebase https://firebase.google.com/docs/firestore/data-model#subcollections

bcho892 (approved):

PR #698, opened 2024-07-28
Fix table on admin availability view

bcho892:
image

Had some issues with the way the start and end date were being checked - also memoized the data


asun555 (approved):
LGTM

PR #701, opened 2024-07-28
change transforms

bcho892:
Avoided using the transform on `svg` as safari does not support SVG2. Instead put the transform on the `path` element:

https://stackoverflow.com/a/78034640

No functionality has been broken on the table/sign up forms


eddieyw (approved):
👍

PR #705, opened 2024-07-28
allow clicking table row to invoke callback

bcho892:
rowcallback

To make use of this you will need to pass in the row identifier, like documented in the `@example`


eddieyw (approved):
Approved

PR #707, opened 2024-07-28
fix booking creation if user already has a booking

bcho892:
before the check meant that users couldn't be added if they had a booking, but that was not specific to the range.


asun555 (approved):
LGTM

PR #708, opened 2024-07-28
610 frontend sort the membership prices before returning thme

asun555:


bcho892 (approved):
"bert"

client/src/services/AppData/AppDataService.test.ts
```diff
@@ -0,0 +1,62 @@
+import { MembershipTypes } from "@/models/Payment"
+
+const membershipOrder = {
```
bcho892, line 3: nit: 
```suggestion
const membershipOrder: {[a: MembershipTypes]: number} = {
```

bcho892: or using `Map` could be a better data structure

PR #710, opened 2024-07-28
remove hard redirect on admin pages to improve UX

bcho892:
Right now the redirects does not play nicely with SSG, as the auth state seems to reset when the page is lazy-loaded.

We assume the only people accessing this route are admins, and if its is not, then an infinite loader is an acceptable experience.


asun555 (approved):
LGTM

PR #712, opened 2024-07-28
add indicators for statuses on admin booking creation

bcho892:
Added statuses:

- No results found
- Loading (has users to fetch)

Also fixed the search query not accounting for admins searching for full names


asun555 (approved):
LGTM

PR #714, opened 2024-07-28
Create schema for gallery

tmacsmee:
I've included `year`, `event`, and `location` as tags.
Only `year` is a required field.


yarn.lock
bcho892, line 1: are these changes required? 

I tried reverting it back to the `master` version, running `yarn cache clean` into `yarn` and it works fine again.

server/src/middleware/__generated__/routes.ts
bcho892, line 1: This can be reverted if the `yarn.lock` is updated

yarn.lock
bcho892: How is your yarn version being managed btw? https://github.com/UoaWDCC/uasc-web/wiki/Onboarding

client/sanity/schema.ts
bcho892, line 1: not needed in this PR, but to add to studio will need to edit `client\sanity.config.ts`

bcho892 (commented):
LGTM, just some minor housekeeping

bcho892 (approved):

PR #699, opened 2024-07-28
add register button to login page

bcho892:
registerbutton

PR #715, opened 2024-07-29
wrap button in link to prevent submit handler being called

bcho892:
Fixes issue introduced in https://github.com/UoaWDCC/uasc-web/pull/699#issuecomment-2254719594


jeffplays2005 (approved):
LGTM! Works on testing too.

PR #716, opened 2024-07-29
Changed default message for backend server

jeffplays2005:


server/src/index.ts
```diff
@@ -42,15 +42,17 @@ app.use("/api-docs", swaggerUi.serve, async (_req: Request, res: Response) => {
...
 })
 
 RegisterRoutes(app)
 
 const port = process.env.PORT || 8000
 
 const _app = app.listen(port, () => {
-  console.log(`Example app listening on port ${port}`)
+  console.log(
+    `UASC backend server listening on port ${port}. http://localhost:${port}`
```
bcho892, line 54: this is not actually true - what if it runs on the prod server

jeffplays2005: True, removed that statement!

bcho892 (approved):

PR #717, opened 2024-07-29
add the lodge info schema to sanity

bcho892:
image

We will have to create a renderer for rich text blocks, likely involving one of the following:

- https://github.com/sanity-io/block-content-to-react/issues/26
- https://github.com/portabletext/react-portabletext


client/src/models/sanity/LodgeInfo/Utils.ts
bcho892, line 1: We need to add the types after we get a renderer like https://github.com/portabletext/react-portabletext

asun555 (approved):
LGTM

PR #719, opened 2024-07-29
Created page for blocking non-members from bookings

Kartik-M24:
This isn't complete yet just had a question cause I couldn't center the div


Kartik-M24:
image
Small mobile view

image
Large mobile view

image
Tablet view

client/src/components/composite/Bookings/NonLoggedInBookingsPage/NonLoggedInBookingsPage.tsx
bcho892, line 1: I'd suggest just making these changes in 

`client\src\components\generic\SignUpNotif\SignUpNotif.tsx`

Kartik-M24:
Also, I just wanted to double-check: if this is a generic component, should I change it so that it's not full length and just have it so it has a bit of padding on all sides?

client/src/components/generic/SignUpNotif/SignUpNotif.tsx
```diff
@@ -21,21 +22,30 @@ export const SignUpNotif = ({ signedIn }: ISignUpNotif) => {
...
-        Only paid members can access bookings. Please sign up or wait until your
-        membership payment has been proccessed.
-      </h4>
-      <span className="flex w-full max-w-80 gap-2">
-        {!signedIn && <Button onClick={() => goToLogin()}>Log in</Button>}
-        <Button onClick={() => goToRegister()}>
-          {signedIn ? "Join Now" : "Sign up"}
-        </Button>
-      </span>
+    <div className="bg-gray-1 absolute inset-0 flex h-screen w-full justify-center overflow-hidden bg-opacity-70 px-4 py-2">
```
bcho892, line 25: can we not have this absolute, i.e make it the same style as the old one, just with the additional lock?

server/package.json
```diff
@@ -12,6 +12,7 @@
     "helmet": "^7.1.0",
     "nodemailer": "^6.9.14",
     "pug": "^3.0.3",
+    "server": "file:",
```
bcho892, line 15: needs to be reverted

yarn.lock
bcho892, line 1: should be reverted

client/package.json
bcho892, line 1: revert this

bcho892 (commented):
if you check it on actual app you'll see that the thing isn't centered, it would be better for you to let it stretch the whole width:

image

Kartik-M24:
The prechecks said to run yarn, after doing so the only file that updates is yarn.lock which I had deleted, what do you recommend I do?

bcho892:
> The prechecks said to run yarn, after doing so the only file that updates is yarn.lock which I had deleted, what do you recommend I do?

You shouldn't delete `yarn.lock` as that tells yarn what it needs to download.

Instead what you should do is revert it to the master version (make sure you have the latest master) with `git checkout master yarn.lock`

server/package.json
bcho892, line 1: revert changes here

client/package.json
bcho892, line 1: revert

yarn.lock
bcho892, line 1: revert

bcho892 (commented):
image

doesn't appear to match design (not centered etc)

client/src/components/generic/SignUpNotif/SignUpNotif.tsx
```diff
@@ -22,15 +23,20 @@ export const SignUpNotif = ({ signedIn }: ISignUpNotif) => {
 
   return (
     <div
-      className="text-dark-blue-100 border-gray-3 flex flex-col items-center justify-center 
-    gap-4 rounded-md border bg-white py-8"
+      className="border-gray-3 flex max-h-[402px] max-w-[570px] flex-col items-center justify-center
```
bcho892, line 26: the problem with constraining in here means that it is not reusable, generally the way to do this is allow containers to stretch max width/height, and use a parent container to constrain it

Kartik-M24: I see that fixed the issue; thanks for the tip.

PR #729, opened 2024-07-30
invalidate booking availabilities after adding user

bcho892:
Needed to invalidate the key for booking availability


asun555 (approved):
LGTM

client/src/services/Admin/AdminMutations.ts
```diff
@@ -129,6 +129,10 @@ export function useAddUserToBookingMutation() {
       queryClient.invalidateQueries({
         queryKey: [ALL_BOOKINGS_BETWEEN_RANGE_QUERY]
       })
+
+      queryClient.invalidateQueries({
```
bcho892, line 133: Each query combines the array elements soust do separately

PR #739, opened 2024-08-04
create paginated endpoint to fetch all booking history

bcho892:
Have only the one dealing with date ranges, the one querying by date range should be implemented later - may not even need it due to it being less intuitive to implement on the front-end


asun555 (approved):
LGTM

PR #750, opened 2024-08-07
Fix videos on ios

bcho892:
Made sure that we use the `playInline` attribute, and also took into consideration that `webm` may not be supported


jeffplays2005 (approved):
Tested and works fine on safari and iPhone safari

client/public/videos/hero-background.mp4
bcho892, line 1: trimmed down to 16s

PR #756, opened 2024-08-11
make date more human readable

ganznz:


server/src/service-layer/controllers/AdminController.ts
```diff
@@ -417,10 +417,20 @@ export class AdminController extends Controller {
...
           authService.getMembershipType(customClaims)
 
+        const formattedDateJoined = metadata
+          ? new Date(metadata.creationTime).toLocaleString("en-US", {
+              year: "numeric",
+              month: "short",
+              day: "numeric",
+              hour: "2-digit",
+              minute: "2-digit"
+            })
```
bcho892, line 427: prefer to have this just give a date, in `dd/mm/yyyy` format

PR #762, opened 2024-08-18
Sort users by their name, and use 1 indexing on booking view

bcho892:
image


asun555 (approved):
LGTM

bcho892:
forgot to take into account capital letters... will find a better string compare method

Issue #631, opened 2024-07-13
[FRONTEND] Connect the policy to sanity


bcho892:
**Is your feature request related to a problem? Please describe.**

Right now policy is hard-coded, we should make a schema that uses https://www.sanity.io/docs/block-type

Refer to https://github.com/UoaWDCC/uasc-web/wiki/Sanity-Studio-Development

We need the following sections:

![Image](https://github.com/user-attachments/assets/9cc6f75a-e1da-4ec8-9aaa-752b6d87326d)

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #763, opened 2024-08-18
660 frontend put lodge schemas on bookings page

bcho892:
Added new dependency https://www.npmjs.com/package/@portabletext/react to help render the text blocks (typed by `PortableTextBlock`. This should be used for any schema that requires rich text in the future - may be relevant for #631. 

also deprecated `ProtectedCreateBookingSection` as now `BookingInformationAndCreation` should be used instead

image


asun555 (approved):
LGTM

PR #764, opened 2024-08-19
Fix build errors due to undefined access

bcho892:
image

This fixes the error in case nothing is fetched from sanity


asun555 (approved):
LGTM

PR #769, opened 2024-08-28
768 frontend display total bookings on admin view

bcho892:
image

We now display the total number of bookings next to the date range


asun555 (approved):
LGTM

PR #770, opened 2024-08-28
Restyle lodge booking page to be more robust

bcho892:
Right now the problem with the current one is that it does not play well on smaller screens, and the photo gallery is pretty hard to use. To solve this it makes more sense to keep styling simpler.

- Simplified gallery to just horizontally scroll
- Only use one column
- Add button to skip the lodge information


asun555 (approved):
LGTM

PR #774, opened 2024-09-14
Add getLodgePrices endpoint to fetch lodge prices

jeffplays2005:


jeffplays2005:
Can't perform testing as don't have access to Stripe key

jeffplays2005:
Conducted some manual testing 
image

server/src/service-layer/controllers/PaymentController.ts
```diff
@@ -107,6 +109,58 @@ export class PaymentController extends Controller {
...
+        switch (lodgeType) {
+          case LodgePricingTypeValues.SingleFridayOrSaturday: {
+            name = LodgePricingTypeValues.SingleFridayOrSaturday
+            break
+          }
+          case LodgePricingTypeValues.Normal: {
+            name = LodgePricingTypeValues.Normal
+            break
+          }
+        }
```
bcho892, line 140: nit: i know most logic here is copy pasted but this can be inlined

bcho892 (approved):

server/src/service-layer/response-models/PaymentResponse.ts
```diff
@@ -31,6 +34,18 @@ export interface MembershipStripeProductResponse extends CommonResponse {
...
+// Make a data shape matching to the expected response from Stripe API
+export interface LodgeStripeProductResponse extends CommonResponse {
+  data?: {
+    productId: string
+    name: LodgePricingTypeValues
+    description?: string
+    discount: boolean
+    displayPrice: string
+    originalPrice?: string
+  }[]
```
bcho892, line 46: nit: documentation of each field

server/src/service-layer/controllers/PaymentController.ts
bcho892: what i mean is just set the `name: lodgeType`

PR #776, opened 2024-09-14
Set up SSE (Server sent events) for event signup counts

jeffplays2005:
Created the getActiveEvents to fetch present events


jeffplays2005:
Note that this could probably do better than using just a Promise<void> statement

image

Manual tests conducted:
image

(Note that the listener also correctly ends when the connection drops)

server/src/data-layer/services/EventService.ts
```diff
@@ -25,6 +26,24 @@ class EventService {
     return result.data()
   }
 
+  /**
+   * Fetches all events that have a start date in the future.
+   * Note that "active" means any event that is still upcoming, or "ready" to signup.
```
bcho892, line 31: does that mean we need to check the `end_date` hasn't expired?

bcho892: that raises another question, we might need an extra `event_date` for when the event actually is

bcho892:
also PR name

server/src/data-layer/services/EventService.ts
jeffplays2005: > does that mean we need to check the `end_date` hasn't expired?

No, not necessarily. The `start_date` variable is supposedly ≥ `end_date` so essentially any `start_date` that is ≥ today would be a "scheduled" event. 

> that raises another question, we might need an extra `event_date` for when the event actually is

But couldn't events be lasting several days?

bcho892: I think my reasoning is that the signup for the event is different from the actual event date

jeffplays2005: > I think my reasoning is that the signup for the event is different from the actual event date

Ahhh right, I supposed that an event start_date & end_date was the actual planned dates for the signups and that they close reservations from the point of start_date

bcho892 (approved):

PR #785, opened 2024-09-16
backend create endpoint for admin to create event and also to fetch all events

bcho892:
# Admin event creation and paginated fetching events

created two endpoints:

- `GET /events`
   - Allows users to do cursor based pagination on the latest events
- `POST /admin/events`
   - Simple **create** operation

Note the changes used for sorting timestamps in `server/src/data-layer/services/EventService.ts` need to be implemented in `BookingHistory` as the are currently not being sorted properly :skull:.

### DIFF
--- a/server/src/data-layer/services/EventService.ts
+++ b/server/src/data-layer/services/EventService.ts
@@ -118,6 +118,46 @@ class EventService {
       .doc(reservationId)
       .delete()
   }
+
+  public async getEventSnapshot(id: string) {
+    return await FirestoreCollections.events.doc(id).get()
+  }
+
+  /**
+   * Returns a list of the latest events. Note that "latest" means events with the latest start date.
+   *
+   * @param limit how many events to fetch
+   * @param startAfter snapshot of document which was the last cursor - should
+   * fetch using the helper defined as {@link getEventSnapshot}
+   * @returns an object providing the id of the next snapshot and list of events
+   */
+  public async getAllEvents(
+    limit: number,
+    startAfter?: FirebaseFirestore.DocumentSnapshot<
+      Event,
+      FirebaseFirestore.DocumentData
+    >
+  ) {
+    let query = FirestoreCollections.events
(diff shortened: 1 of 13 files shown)


server/src/data-layer/services/EventService.ts
```diff
@@ -158,6 +158,46 @@ class EventService {
       .doc(reservationId)
       .delete()
   }
+
+  public async getEventSnapshot(id: string) {
+    return await FirestoreCollections.events.doc(id).get()
+  }
```
jeffplays2005, line 164: Nit: do we need documenting?

server/src/service-layer/controllers/AdminController.ts
```diff
@@ -692,4 +694,20 @@ export class AdminController extends Controller {
...
+  public async createNewEvent(@Body() body: CreateEventBody) {
+    try {
+      const eventService = new EventService()
+      await eventService.createEvent(body.data)
+
+      this.setStatus(201)
+    } catch {
+      this.setStatus(500)
+    }
+  }
```
jeffplays2005, line 712: Just for future, should we use GET, POST, PATCH methods to indicate for fetching, creating, editing in the `/admin/events/ endpoint`?

jeffplays2005: Was also thinking of creating another ticket to address these changes as a few endpoints have some endpoints that could use the three methods instead of explicitly stating "create".

bcho892: 👍

bcho892: yeah actually probably a good idea to do it now

bcho892: we should write a wiki article enforcing this?

jeffplays2005 (approved):
:p lgtm!

asun555 (approved):
LGTM

PR #787, opened 2024-09-30
Rewrite SSE endpoint to return map of reservations per event ID

jeffplays2005:
I've changed the method in the controller to use the modified method.

Also changed EventService to return a Record instead of just a number.

Updated tests

### DIFF
--- a/server/src/data-layer/services/EventService.test.ts
+++ b/server/src/data-layer/services/EventService.test.ts
@@ -246,8 +246,8 @@ describe("EventService integration tests", () => {
       await eventService.addReservation(newEvent.id, reservation1)
       await eventService.addReservation(newEvent.id, reservation2)
 
-      const count = await eventService.getActiveReservationsCount()
-      expect(count).toBe(2)
+      const eventCounts = await eventService.getActiveReservationsCount()
+      expect(eventCounts).toStrictEqual({ [newEvent.id]: 2 })
     })
 
     it("Should get all event reservations", async () => {
--- a/server/src/data-layer/services/EventService.ts
+++ b/server/src/data-layer/services/EventService.ts
@@ -116,18 +116,18 @@ class EventService {
 
   /**
    * Used for the SSE feature to display the total number of active event reservations.
-   * @returns the total number of active event reservations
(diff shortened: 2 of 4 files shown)


jeffplays2005:
Note that this endpoint still doesn't have tests due to the nature of it having an open socket.

I've conducted some manual tests but had an inability to add an event of the correct data structure which results in problems with fetching all the events with a later date.

See the images attached below:
The following image depicts that the endpoint returns an object instead of a number now:
Screenshot 2024-10-01 at 12 22 53 PM

Upon adding an event with the admin endpoint, the event doesn't seem to be fetched (this is perhaps due to some firestore timestamp issue again)

Note that the first dump was of just fetching regular events(this included the new event I created) but upon filtering with dates in Firestone, it only returns the empty lists after. 

Screenshot 2024-10-01 at 12 24 32 PM

<details>
<summary>JSON event that was tested</summary>

```json
{
    "data": {
        "title": "string",
        "description": "string",
        "image_url": "string",
        "location": "string",
        "physical_start_date": {
            "seconds": 1738321200,
            "nanoseconds": 0
        },
        "start_date": {
            "seconds": 1738321200,
            "nanoseconds": 0
        },
        "end_date": {
            "seconds": 1738321200,
            "nanoseconds": 0
        },
        "physical_end_date": {
            "seconds": 1738321200,
            "nanoseconds": 0
        },
        "max_occupancy": 30
    }
}
```

</details>

bcho892:
do we not have confirmation that the endpoint can provide signups for each event id?

jeffplays2005:
> do we not have confirmation that the endpoint can provide signups for each event id?

I've tested the function in the service tests and this endpoint essentially just returns that. I'll try and get some jest tests set up if it's doable.

It isn't working when I manually add a payload to the admin/events POST method

bcho892 (approved):

jeffplays2005:
Screenshot 2024-10-01 at 11 33 43 PM
Works properly now, can assume that the counts will be accurate as the service methods have been tested and the SSE just returns the tested service methods.

The changes on the admin endpoint should convert the raw timestamps to firestore timestamps

PR #791, opened 2024-10-04
783 backend write a cron job to automatically update a google sheet with the members

jeffplays2005:
Still semi-draft...

*. Update dependencies in server workspace
* Create new script in `package.json`
* Created new cron job(note that the cron is currently disabled and has workflow-dispatch for now).
* Create `fetchUsers` method that fetches users from the backend
* Create `getAllUsers` that fetches all the users from the backend using `fetchUsers`
* Create `authenticateGoogle` method that returns a google auth client
  * Note that documentation will come soon...
* Create `appendToGoogleSheet` and `clearSheet`
  * `appendToGoogleSheet` takes in the logs and auth client and updates the google sheet rows
  * `clearSheet` clears the google sheet so that `appendToGoogleSheet` doesn't overlap everything

>[!Note] 
The logic is still somewhat rough and could have improvements. Quite a lot of duplicate code and possibly better ways to update the google sheet instead of just appending and clearing every time.


server/package.json
```diff
@@ -9,6 +9,7 @@
     "dotenv": "^16.4.5",
     "express": "^4.18.2",
     "firebase-admin": "^12.0.0",
+    "googleapis": "^144.0.0",
```
bcho892, line 12: should this be a dev dependency?

server/tooling/update-google-sheet-members.ts
```diff
@@ -0,0 +1,177 @@
...
+    console.error("Failed to clear the Google Sheet", err)
+  }
+}
+
+if (args.length === 0) {
+  console.log("Token required to update google sheet members.")
+} else {
+  const token = args[0]
+
+  getAllUsers(token)
```
bcho892, line 131: this should be done synchronously

server/tooling/update-google-sheet-members.ts
```diff
@@ -0,0 +1,177 @@
...
+    console.log(`Cleared google sheet.`)
+  } catch (err) {
+    console.error("Failed to clear the Google Sheet", err)
+  }
+}
+
+if (args.length === 0) {
+  console.log("Token required to update google sheet members.")
+} else {
+  const token = args[0]
```
bcho892, line 129: its better to generate the token in this script rather than the workflow

.github/workflows/update_google_sheet_members.yml
```diff
@@ -0,0 +1,34 @@
+name: Update the google sheet with members
+
+on:
+  workflow_dispatch:
+  # schedule:
+  # Updates every 00:00 every day
+  # - cron: "0 0 * * *"
+
+env:
+  NEXT_PUBLIC_BACKEND_BASE_URL: ${{secrets.VITE_BACKEND_BASE_URL}}
```
bcho892, line 10: you are missing the service account env variable

server/tooling/update-google-sheet-members.ts
bcho892: https://firebase.google.com/docs/auth/admin/create-custom-tokens#create_custom_tokens_using_the_firebase_admin_sdk, and also refer to how a token is made in https://github.com/UoaWDCC/uasc-web/blob/master/server/tooling/login-prod.ts#L34

bcho892 (changes requested):

jeffplays2005:
Added some documentation in here:

https://github.com/UoaWDCC/uasc-web/wiki/Deployment-Information#google-sheet-members-action

server/tooling/update-google-sheet-members.ts
```diff
@@ -0,0 +1,225 @@
...
+ */
+async function getAllUsers(token: string): Promise<CombinedUserData[]> {
+  const allUsers: CombinedUserData[] = []
+  let fetchedUsers = await fetchUsers(token)
+  allUsers.push(...fetchedUsers.data)
+  while (fetchedUsers.cursor) {
+    fetchedUsers = await fetchUsers(token, fetchedUsers.cursor)
+    allUsers.push(...fetchedUsers.data)
+  }
+  return allUsers
```
bcho892, line 79: do we fetch all users and not check if they're members?

bcho892 (commented):
have you tested the function by running `ts-node` locally?

jeffplays2005:
> have you tested the function by running `ts-node` locally?
>
> do we fetch all users and not check if they're members?
image

jeffplays2005:
Note that I'll need to document more on the wiki as the `GOOGLE_SERVICE_ACCOUNT_JSON` doesn't work for this... the `GOOGLE_SERVICE_ACCOUNT_JSON` only has access for firebase I believe although this use case needs an account json for sheets

bcho892 (approved):

jeffplays2005:
For documentation in this branch:

Updated secrets

Screenshot 2024-10-05 at 11 31 35 PM

PR #798, opened 2024-10-06
Add google forms link as field to event and clarify fields

bcho892:
This is because we are removing the firestore signups for events in #797. This field is also optional so the event can be created pretty far out from the sign up date.

Renamed the `start_date` and `end_date` to `sign_up_start_date` and `sign_up_end_date` to fix the bad naming


asun555 (approved):
LGTM

PR #799, opened 2024-10-06
796 backend events endpoint only include forms url at given time

AzizPatel786:


server/src/middleware/tests/EventController.test.ts
```diff
@@ -62,6 +73,33 @@ describe("EventController endpoint tests", () => {
...
+    it("should not include google_forms_link if event is not within 1 minute", async () => {
+      await eventService.createEvent(event3)
+
+      const res = await request.get("/events").send()
+
+      expect(res.body.data).toContainEqual(
+        expect.not.objectContaining({
+          google_forms_link: expect.not.stringContaining(
+            "https://random.com/event3"
+          )
```
bcho892, line 99: you should instead check if its undefined, otherwise this test breaks easily

bcho892: if its a different string the behaviour is still "incorrect"

server/src/service-layer/controllers/EventController.ts
```diff
@@ -97,6 +97,19 @@ export class EventController extends Controller {
...
+
+      res.events.forEach((event) => {
+        const eventStartTime = event.physical_start_date
+        const timeDifference =
+          eventStartTime.toMillis() - currentTime.toMillis()
+
+        // 1 minute (60000 milliseconds)
+        if (timeDifference > 60000) {
+          delete event.google_forms_link
+        }
```
bcho892, line 110: nit: magic number

bcho892 (changes requested):
LGTM, test is just a bit brittle right now

bcho892 (approved):
Sorry I missed this

PR #800, opened 2024-10-07
797 backend delete all code relating to event sign ups

jeffplays2005:
* Deleted event signup endpoint
* Deleted SSE endpoint
* Removed relevant endpoint tests
* Remove EventReservation service methods and tests
* Remove event reservations from firebase and subcollection
  * Note that `Event.max_occupancy` is kept as this could still be useful.
  * Also note that `FirestoreSubcollections.ts` was kept as this could be used in the future.
  * Also removed the event request for signup


bcho892 (approved):
RIP

Issue #779, opened 2024-09-15
[FRONTEND] Create component edit view for event


bcho892:
**Is your feature request related to a problem? Please describe.**

Needs to contain the fields admins require to edit the event.

These are defined in https://github.com/UoaWDCC/uasc-web/wiki/Firestore-Architecture#event_slot-collection

The design can be something simple, i.e

![Image](https://github.com/user-attachments/assets/a026e343-f5da-4e97-8d4c-81c9f543035c)

>[!IMPORTANT]
>Only a working storybook is required for this ticket

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

Issue #803, opened 2024-10-09
[FRONTEND] Add CDN over firebase storage


bcho892:
**Is your feature request related to a problem? Please describe.**

For the event images, currently they are being served _raw_,

look into this:

https://www.anothermadworld.com/how-to-put-a-cdn-in-front-of-firebase-cloud-storage/

**BEFORE MERGING**

- [ ] Storybooks created where possible
- [ ] Tested with [nock](https://tanstack.com/query/v3/docs/framework/react/guides/testing) if data fetching
- [ ] Tests written for critical interactions
- [ ] PR Reviewed (For non-trivial changes)
- [ ] Changes tested after rebasing on master or merging in master (_hint_: `git fetch origin master:master`, then `git rebase master` or `git merge master`)
- [ ] All required PR checks passing

PR #804, opened 2024-10-09
801 frontend add interface for admins to create new events

bcho892:
# Form

image

This component will need to be modified to show slight differences for editing, as it can be reused for editing an event. #779

We can start using this to test the fetching all events endpoint


asun555 (approved):
LGTM

client/src/services/Storage/StorageService.ts
jeffplays2005, line 1: Could possibly use imgur for storage?

jeffplays2005 (approved):
Just a few minor suggestions(I also hope this isn't a safari issue too), any way we could have those date boxes aligned? 

Screenshot 2024-10-09 at 10 57 09 PM

Perhaps some padding adjustments between the two headings?
Screenshot 2024-10-09 at 10 57 58 PM

Diff1:
image
Diff2:
Screenshot 2024-10-09 at 11 00 21 PM


NB: These aren't important changes!

Looks clean otherwise !!

client/src/services/Storage/StorageService.ts
bcho892: i would prefer a free service like that, do you think theres enough information about an API that we could make a wrapper for?

bcho892: likely should be handled in #803

PR #809, opened 2024-10-18
772 frontend create component to render all events

bcho892:
# Event Page (User View)

Created most logic for displaying the events, test by checking for any bad behaviour on the `/events` route.

## Detailed View
image

## Sign up area

image

## Preview events for upcoming and current

>[!NOTE]
>Upcoming events are sorted ascending (so closest to "now"), a past event is defined as the end date elapsing or a day passing from the start date if no end date is given

image

## Past Events

image


client/src/components/generic/Event/EventDetailed/EventDetailed.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <div className="flex w-full flex-col items-center justify-center gap-4">
+        <div className="flex h-fit w-full justify-center border bg-white">
+          <Image
+            src={image}
+            alt="Event cover image"
+            width={500}
+            height={300}
+            className="block h-auto max-h-[400px] w-auto  py-4"
+          />
+        </div>
```
jeffplays2005, line 94: image

Is it intentional that the border doesn't wrap the image and overlaps in a rectangle? Thought just as an improvement, could just have the image centre instead.

client/src/components/generic/Event/EventUtils.ts
```diff
@@ -0,0 +1,211 @@
...
+   */
+  signUpOpen: (signUpOpenDate: Date) => {
+    const currentDate = new Date()
+    // Sign Ups have opened
+    if (signUpOpenDate <= currentDate) {
+      return `Sign ups have opened! (${EventDateFormatting.getRelativeTimeString(signUpOpenDate)})` as const
+    } else {
+      return `Sign ups open ${EventDateFormatting.getRelativeTimeString(signUpOpenDate)} (${EventDateFormatting.shortDate(signUpOpenDate)})` as const
+    }
+  },
```
jeffplays2005, line 24: Super nice to have this although think we can have a date range or something that specifies the ending date of signups?

client/src/components/generic/Event/EventDetailed/EventDetailed.tsx
jeffplays2005, line 94: Just out of curiosity as well, we might need Miguel for some designing here as the image being detached from the main div makes it a bit difficult to navigate, could we instead have the image on the side?

jeffplays2005 (commented):
Just a bunch of nit picks on some designing here and there but everything looks quite solid! 

Also wondering, could we have the events instead being on the same page as /events, but use /events/{id} instead? Just thought it'd make sharing easier

client/src/components/generic/Event/EventUtils.ts
bcho892: the client never close sign ups anyway, so for us we just close it after the event is over

client/src/components/generic/Event/EventDetailed/EventDetailed.tsx
bcho892: That was part of the figma design, I'd argue that on the side would squish the text too much, however it being on the bottom may work 

image

bcho892: the image should ideally fill the container so yes its not intentional, will need to be resolved

bcho892:
> Also wondering, could we have the events instead being on the same page as /events, but use /events/{id} instead? Just thought it'd make sharing easier

Can probably use a query param (i.e `/events?id={id}`), as that's the only way with static export

client/src/components/generic/Event/EventDetailed/EventDetailed.tsx
jeffplays2005: Sounds good

client/src/components/generic/Event/EventUtils.ts
bcho892: but yea if they decide to put it in may be worth

asun555 (approved):
LGTM

zlrkw11 (approved):
clean

client/src/components/generic/Event/EventDetailed/EventDetailed.tsx
bcho892: done

bcho892: done

jeffplays2005 (commented):
One last pick as this is highly unlikely but still possible, is there any default asset that we can use in place if no event cover image is provided?
image
image

Thought it would make the website look more "professional"

bcho892:
> One last pick as this is highly unlikely but still possible, is there any default asset that we can use in place if no event cover 
> Thought it would make the website look more "professional"

https://placehold.co/600x400?text=UASC+Event

jeffplays2005:
> > One last pick as this is highly unlikely but still possible, is there any default asset that we can use in place if no event cover
> > Thought it would make the website look more "professional"
> 
> https://placehold.co/600x400?text=UASC+Event

I tried running this locally although the client doesn't seem to allow for svgs/xml coming from external sources. I believe this is probably 
image

image

bcho892:
You should check the preview link which has the latest changes

jeffplays2005 (approved):

PR #810, opened 2024-10-19
fix double initialistation of firebase when creating storage service

bcho892:
same issue as https://github.com/vercel/next.js/discussions/11351https://github.com/vercel/next.js/discussions/11351, however we just need to not call `getStorage` raw without passing in the firebase app

This was causing the `admin/events` build to sometimes fail


zlrkw11:
lookin crisp

zlrkw11 (approved):
clean

PR #812, opened 2024-10-19
adding regex for .com

zlrkw11:


zlrkw11:
image
- regex specified for .com
- user cannot enter invalid emails and will be notified immediately upon

bcho892 (approved):

PR #815, opened 2024-10-25
allow form to be used for both editing and creating

bcho892:
Added two props to the `AdminEventForm`:

- `isEditingMode`
- `defaultData`

Which allow the form to be used to send edit requests with the `PATCH` endpoint for events

>[!NOTE]
>Also fixed the padding in the Event Preview, because the use of numbers conditionally stopped tailwind from being compiled properly

## Normal Create View

image

## Editing View

image


eddieyw (approved):
approved

PR #817, opened 2024-11-02
Scope get event id endpoint to admins only

jeffplays2005:
* Updated endpoint

* Updated interfaces

* Updated tests

* Generate swagger

PR #818, opened 2024-11-02
816 frontend implement editing functionality for events

bcho892:
Made use of the work in #815, and created the new queries for fetching a single event (waiting for #817) and mutation for editing the event. 

editevent

Have added placeholders for the delete functionality, which an endpoint needs to be made for

Also fixed a logical error in #799 where it was checking the event start date instead of sign up start date


asun555 (approved):
LGTM

PR #824, opened 2024-11-03
Properly refresh the list of events after being refetched

bcho892:
With this change we now assume that all events are invalid and refetch the whole list, note that this is pretty much the same performance as before anyway - but stops the user from accessing out-of-date data.

cacheinvalid


jeffplays2005 (approved):
👍 gw!

PR #829, opened 2024-11-12
828 frontend dynamically fetch lodge prices on frontend

choden-dev:
image

image

Still works, will require documentation if this is to be added to any other features


asun555 (approved):
LGTM

PR #860, opened 2025-06-14
859 Merge CI Script

jeffplays2005:
Merged the following scripts into `CI-pipeline`:

- Prettier
- Eslint
- Tests
- Check builds
- Check code generation

Can be optimised further if build is cached as well (although doesn't make it much better due to the wait of 2 minute build time)...

If using cache: waits for build (2 min), then proceeds with 20 second cache
Not using cache: runs parallel, so reduces the 20 second cache download

Screenshot 2025-06-14 at 5 41 10 PM


jeffplays2005:
@choden-dev will have to temporarily update settings to remove the old stale pending checks for the new `verify` check

choden-dev (approved):

choden-dev:
@jeffplays2005 surely the server and client build should be counted as separate in the pipeline, otherwise LGTM

jeffplays2005:
> @jeffplays2005 surely the server and client build should be counted as separate in the pipeline, otherwise LGTM

Done

.github/workflows/ci-pipeline.yml
```diff
@@ -0,0 +1,206 @@
...
+          NEXT_PUBLIC_BACKEND_BASE_URL: ${{ secrets.VITE_BACKEND_BASE_URL }}
+          NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY: ${{ secrets.TEST_STRIPE_PK }}
+          NEXT_PUBLIC_SANITY_PROJECT_ID: ${{ secrets.STAGING_SANITY_PROJECT_ID }}
+          NEXT_PUBLIC_SANITY_DATASET: staging
+          NEXT_CONFIG_ENV: staging
+
+  build-server:
+    needs: [check_changes, setup]
+    if: ${{ needs.check_changes.outputs.exists == 'true' }}
+    name: Build
```
choden-dev, line 156: sorry, just need to change this to build server for clarity

jeffplays2005: Oh yep just noticed that as well !

PR #861, opened 2025-06-14
852 frontend link google sheets url on admin table

choden-dev:
Encapsulated the logic in the backend, now we have a generic endpoint that can be used to return sensitive URLs to admins only, first we do this with a button on top of the table:

image

## Redirect Endpoint Details

This PR adds a new endpoint for retrieving redirect URLs stored in environment variables. This allows for externally configurable URLs that can be managed without code changes.

### Usage

The endpoint accepts a redirect key and returns the corresponding URL from environment variables:

```
GET /admin/redirect/{redirectKey}
```

- **Parameters**: `redirectKey` - A string identifier matching a value in the `RedirectKeys` enum
- **Authentication**: Requires admin JWT token
- **Returns**: JSON object with either a `url` property containing the URL string, or an `error` property

### Implementation Notes

- The endpoint looks for environment variables with the format `REDIRECT_{KEY}`
- Keys are automatically converted to uppercase and trimmed
- The key must exist in the `RedirectKeys` enum (validation step)
- Current supported keys include `MEMBERS_GOOGLE_SHEET_LINK`
- Returns 404 if the key is not found or not valid
- Returns 500 for unexpected errors

### Example Response

```json
{
  "url": "https://docs.google.com/spreadsheets/d/example"
}
```

This implementation enables administrators to access configured URLs without hard-coding them in the application.


client/src/components/composite/Admin/AdminMemberView/AdminMemberView.tsx
```diff
@@ -202,7 +215,10 @@ export const AdminMemberView = ({
     <div
       className={`w-full ${isUpdating ? "brightness-75" : "brightness-100"}`}
     >
-      <span className="mb-4 mt-6 flex w-full flex-col justify-between sm:flex-row">
+      <Button variant="inverted-default-sm" onClick={handleGoToGoogleForm}>
+        Go to Google Form
+      </Button>
```
jeffplays2005, line 220: nit: is a google sheet i think

jeffplays2005 (approved):
LGTM! Clean implementation

client/src/components/composite/Admin/AdminMemberView/AdminMemberView.tsx
choden-dev: oh good catch lol

PR #921, opened 2025-07-22
fix: use secret for REDIRECT_MEMBERS_GOOGLE_SHEET_LINK in CD pipeline configuration

choden-dev:
# Description

Properly set the environment variable for the google sheet. Right now only planning on populating it in prod

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -178,8 +178,8 @@ jobs:
   verify:
     name: Verify
     needs: [check_changes, lint, prettier, test, codegen, build-client, build-server]
+    if: ${{ needs.check_changes.result == 'success' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pipeline.yml
```diff
@@ -178,8 +178,8 @@ jobs:
   verify:
     name: Verify
     needs: [check_changes, lint, prettier, test, codegen, build-client, build-server]
+    if: ${{ needs.check_changes.result == 'success' }}
     runs-on: ubuntu-latest
-    if: ${{ always() && (needs.check_changes.outputs.exists != 'true' || (needs.lint.result == 'success' && needs.prettier.result == 'success' && needs.test.result == 'success' && needs.codegen.result == 'success' && needs.build-client.result == 'success' && needs.build-server.result == 'success')) }}
```
jeffplays2005, line 182: question: was this intentional? we shouldn't do this as verify will pass without the required checks passing.

choden-dev: I was testing because it was passing as the verify job was skipped after some steps failed (therefore the pipeline thought it passed). We need a way to make sure the verify job fails instead

